PR #18, opened 2015-06-04
Vector2D methods and unit tests, Line2D analog of Line3D and unit tests

mattj23:
I added the following two methods to the Vector2D struct:

* CrossProduct - performs the commonly used 2d 'cross product' equivalent, which returns a positive or negative scalar that would be the z component of the 3D cross product if both vectors were 3D vectors in the z=0 plane

* ProjectOn - projects one vector onto another vector, following the style of Vector3D.ProjectOn

Additionally I added unit tests to cover those methods and a few others, bringing the test coverage up to 85% by branch, 88% by lines of code.

I also added a Line2D struct which follows the example of Line3D, and unit tests to cover all of the math portions (83% by branch, 93% by lines of code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ coverage.*
 *.psess
 *.vsp
 *.vspx
+*.boltdata
 
 # Caches
 _ReSharper*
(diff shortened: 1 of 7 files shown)


cdrnet:
Great, thanks a lot!

JohanLarsson:
Nice, thanks!

PR #20, opened 2015-06-25
Plane updates

EMostafaAli:
* (Added) Plane constructor using 3 Point3d
* (Added) Line and Plane intersection

### DIFF
--- a/src/Spatial/Euclidean/Plane.cs
+++ b/src/Spatial/Euclidean/Plane.cs
@@ -41,6 +41,25 @@ namespace MathNet.Spatial.Euclidean
         }
 
         /// <summary>
+        /// http://www.had2know.com/academics/equation-plane-through-3-points.html
+        /// </summary>
+        /// <param name="p1"></param>
+        /// <param name="p2"></param>
+        /// <param name="p3"></param>
+        public Plane(Point3D p1, Point3D p2, Point3D p3)
+        {
+            if (p1 == p2 || p1 == p3 || p2 == p3)
+            {
+                throw new Exception("Must use three different points");
+            }
+            Vector3D v1 = new Vector3D(p2.X - p1.X, p2.Y - p1.Y, p2.Z - p1.Z);
+            Vector3D v2 = new Vector3D(p3.X - p1.X, p3.Y - p1.Y, p3.Z - p1.Z);
+            this.RootPoint = p1;
+            this.Normal = v1.CrossProduct(v2).Normalize();
+            this.D = -this.RootPoint.ToVector3D().DotProduct(this.Normal);
+        }
+
+        /// <summary>
         /// Creates a Plane from its string representation
         /// </summary>
         /// <param name="s">The string representation of the Plane</param>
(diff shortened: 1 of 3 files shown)


JohanLarsson:
Nice & thank you!
A couple of more tests are perhaps needed. What do you say @cdrnet should more tests be added before merging?

src/Spatial/Euclidean/Plane.cs
```diff
@@ -41,6 +41,25 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+        /// <param name="p1"></param>
+        /// <param name="p2"></param>
+        /// <param name="p3"></param>
+        public Plane(Point3D p1, Point3D p2, Point3D p3)
+        {
+            if (p1 == p2 || p1 == p3 || p2 == p3)
+            {
+                throw new Exception("Must use three different points");
+            }
+            Vector3D v1 = new Vector3D(p2.X - p1.X, p2.Y - p1.Y, p2.Z - p1.Z);
```
JohanLarsson: Checking that the points are not on a line makes sense here.

src/Spatial/Euclidean/Plane.cs
```diff
@@ -41,6 +41,25 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+        /// http://www.had2know.com/academics/equation-plane-through-3-points.html
+        /// </summary>
+        /// <param name="p1"></param>
+        /// <param name="p2"></param>
+        /// <param name="p3"></param>
+        public Plane(Point3D p1, Point3D p2, Point3D p3)
+        {
+            if (p1 == p2 || p1 == p3 || p2 == p3)
+            {
+                throw new Exception("Must use three different points");
```
JohanLarsson: ArgumentException is probably better.

src/Spatial/Euclidean/Plane.cs
```diff
@@ -182,6 +201,36 @@ public Ray3D IntersectionWith(Plane intersectingPlane, double tolerance = float.
...
+        /// <param name="tolerance"></param>
+        /// <returns>Intersection Point or null</returns>
+        public Point3D? IntersectionWith(Line3D line, double tolerance = float.Epsilon)
+        {
+            if (line.Direction.IsParallelTo(this.Normal)) //either parallel or lies in the plane
+            {
+                Point3D projectedPoint = this.Project(line.StartPoint, line.Direction);
+                if (projectedPoint == line.StartPoint) //Line lies in the plane
+                {
+                    throw new Exception("Line lies in the plane"); //Not sure what should be done here
```
JohanLarsson: Tricky spot, not sure either. ArgumentException or InvalidOperationException

EMostafaAli: For now,  I will go with InvalidOperationException.

src/Spatial/Euclidean/Plane.cs
EMostafaAli: Yeah, I overlooked 'points on a line'. I will add this check

JohanLarsson:
Ok, I'll try to remember to add some tests later. Thank you for the contribution!
Also, spend som time learning nUnit, it is a useful tool :)

PR #22, opened 2015-07-02
Fix the check for collinear points in Plane constructor

Maghoumi:
Currently, if 3 points are passed to the Plane constructor and the first point is between the other two, the col-linear check in the constructor fails (UnitVector3D class will instead throw the exception). An example is provided below:

Point3D p1 = new Point3D(0, 0, 0);
Point3D p2 = new Point3D(1, 0, 0);
Point3D p3 = new Point3D(2, 0, 0);
Plane p = new Plane(p2, p1, p3);

My fix is based on the magnitude of the cross product of the vectors formed by the 3 points.


JohanLarsson:
Yeah, that is nicer. Thank you.

PR #23, opened 2015-07-21
Adding some extra methods to quaternion

djowatts:


djowatts:
I had these as extension methods in my own project and found myself using them an awful lot. It may be quite specific for my cause, but others may find these and similar things useful.

JohanLarsson:
Thank you!

PR #27, opened 2016-02-14
C# quaternion implementation

MaLiN2223:
[THIS PULL IS NOT FULLY READY YET]

Since C# implementation is not needed in Numerics I am creating it here.
Moved from [it](https://github.com/mathnet/mathnet-numerics/pull/376) 
Things to do :

Actual Code : 
- [ ] Rotations
- [ ] More robust division (possible overflow)
- [ ] Create more robust norm (possible overflow)
- [ ] Create quaternion from angle
- [ ] Create quaternion from matrix 
- [ ] Two rotation concatenation [source] (http://mathworld.wolfram.com/Quaternion.html)

--
Testing
- [ ] DotProduct
- [ ] Inversion
- [ ] Equality, operator ==
- [ ] ToString()
- [ ] Norm
- [ ] Conjugation
- [ ] operator* for floats 
- [ ] operator/
- [ ] Rotations
- [ ] Constructors

If you have any more ideas what to add just post it - I will handle it.
If you want to contribute :  [here is my git repo](https://github.com/MaLiN2223/mathnet-spatial)

### DIFF
--- a/src/Spatial/Euclidean/Quaternion.cs
+++ b/src/Spatial/Euclidean/Quaternion.cs
@@ -31,12 +31,22 @@ namespace MathNet.Spatial.Euclidean
         }
         #endregion
         #region operations
-
+        /// <summary> 
+        /// Squared norm of a quaternion
+        /// </summary>
+        /// <remarks>
+        /// http://mathworld.wolfram.com/QuaternionNorm.html
+        /// </remarks>
         public float NormSquared
         {
             get { return a1 * a1 + a2 * a2 + a3 * a3 + a4 * a4; }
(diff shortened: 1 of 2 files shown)


cdrnet:
Thanks for continuing this work here in Spatial!

cdrnet:
I'm somewhat confused by the first commit. It looks like the implementation was completely replaced? Was this intentional? The existing implementation had a few issues that should be addressed, but was quite powerful already.

Tip: you may want to work on a branch and open the pull request from there instead of master, or you'll forever link your master to this pull request.

PR #28, opened 2016-02-14
New quaternions.

MaLiN2223:
I hope I did better than previous time.
What I did : 
Changes : 
- **New Interfaces**:  
	IEquatable<Quaternion> and IFormattable
- **Added constructors** :  
	public Quaternion(Vector v)  
	public Quaternion(DenseVector v)
- **Changed constructors** :  
	public static Quaternion From4Vector(DenseVector v) (changed implementation for consistency with constructor Quaternion(double real, double imagX, double imagY, double imagZ))
- **Added proporties** :  
	public static readonly Quaternion One;  
	public static readonly Quaternion Zero;
- **Changed proporties**:   
	public double Abs (changed name to NormSquared - more accurate) 
- **Changed Fields** :  
	instead of norm and abs now we have normSquared and norm 
- **Changed static methdos** :   
	private static double ToNorm(double real, double imagX, double imagY, double imagZ) (explicit private modifier)  
	private static Quaternion ToUnitQuaternion(double real, double imagX, double imagY, double imagZ)(explicit private modifier)  
	public static Quaternion operator /(Quaternion q1, Quaternion q2) (now is more robust)
- **Added static methods**:   
	public static Quaternion operator +(double d,Quaternion q1) (because of implicit conversion - I dont want it to get buggy)  
	public static Quaternion operator -(double d,Quaternion q1)  
	public static Quaternion operator *(double d,Quaternion q1)  
	public static Quaternion operator /(double d,Quaternion q1)
- **Changed instance methods** :  
	public Quaternion Scalar() (changed to property)  
	public Quaternion Vector() (changed to property)  
	public Quaternion UnitVector() (changed to property)  
	public Quaternion Sign() (changed to proeprty)  
	public Quaternion Inverse() (changed to property + letter 'd' at the end)
- **Added instance methods**:   
	public bool Equals(Quaternion other)
	public string ToString(string format, IFormatProvider formatProvider)
- **Removed instance methods** :   
	public Quaternion Add(double)
	public Quaternion Add(Quaternion)
 	public Quaternion Subtract(double r) 
       public Quaternion Subtract(Quaternion q) 
       (etc. for API consistency - from now rely on operators)

Still to do : 
Actual Code
- [ ] ~~More robust norm calculation~~
- [x] Comments
- [x] Remove fields for norm, norm squared etc.

--
Tests: 
- [x] Inversion
- [x] Equality, operator ==
- [x] Operators
- [x] Negation
- [x] ToString()
- [x] Norm
- [x] Conjugation
- [x] Rotations
- [x] Constructors
- [x] Argument
- [x] UnitQuaternions
- [x] Normalization 
- [x] Log, Ln , Pow,Exp,Sqrt,Sqr 

Current tests coverage : 83%


cdrnet:
Looking great now, thanks!

src/Spatial/Euclidean/Quaternion.cs
```diff
@@ -25,53 +24,62 @@ public Quaternion(double real, double imagX, double imagY, double imagZ)
             _y = imagY;
             _z = imagZ;
             _w = real;
-            _norm = ToNorm(real, imagX, imagY, imagZ);
-            _abs = Math.Sqrt(_norm);
-            _arg = Math.Acos(real/_abs);
+            _normSquared = ToNorm(real, imagX, imagY, imagZ);
+            _norm = Math.Sqrt(_normSquared); //TODO : robust _norm calculating
+            _arg = Math.Acos(real / _normSquared);
```
JohanLarsson: Should these be precomputed? Makes the struct bigger.

src/Spatial/Euclidean/Quaternion.cs
```diff
@@ -310,15 +334,28 @@ public Quaternion RotateUnitQuaternion(Quaternion unitQuaternion)
...
+            return new Quaternion(q1.Real * d, q1.ImagX * d, q1.ImagY * d, q1.ImagZ * d);
+        }
+        /// <summary>
+        /// Divide a quaternion by a quaternion.
+        /// </summary> 
+        public static Quaternion operator /(Quaternion q, Quaternion r)
         {
-            return q1.Divide(q2);
+            double normSquared = r.NormSquared;
+            var t0 = (r._w * q._w - r._x * q._x - r._y * q._y + r._z * q._z) / normSquared;
```
JohanLarsson: Formatting: we have used no space for a*b and space for a + b. Reads nicely imo.

JohanLarsson:
Great with a PR including tests!

src/Spatial/Euclidean/Quaternion.cs
cdrnet: No, they should not (my fault, bad design idea back in old Iridium)

src/Spatial/Euclidean/Quaternion.cs
MaLiN2223: I will change it asap.

src/Spatial/Euclidean/Quaternion.cs
MaLiN2223: Well, how about doing something like this : 
We store those three values in variables after they are calculated by class user.
I am thinking about storing because for example _normSquared is hidden in 

- bool IsUnitQuaternion
- Quaternion Inversed
- Quaternion Ln()

Thus if we create object Quaternion and try to calculate those tree we would calculare it three times!
What I am thinking of calculates it only one : 
Just do backing field for property NormSquared :
```
public double NormSquared{
    get{
        if(normCalculated){ return _normSquared;}
        calculate norm and norm squared
        normCalculated = true
    }
}
```

Or even 
`if(_norm<0) calculate norm and norm squared`
because norm is always >=0

cdrnet: I would just recompute it on demand, and accepting that in very few cases it may be calculated twice.

MaLiN2223:
@cdrnet 
Yet another question rises :
How do we define Quaternion/ 0.0 and similar calculations?

cdrnet:
1. If unclear I'd make it consistent with Complex and Double, hence `q(0,0)/0 -> NaN`, `q(a,b)/0 when (a,b)!=(0,0) -> PosInfinity`.

src/Spatial/Euclidean/Quaternion.cs
```diff
@@ -152,7 +140,7 @@ public bool IsUnitQuaternion
         /// </summary>
         public bool IsRotation
```
cdrnet: Maybe we can actually drop this one, seems to be redundant with IsUnitQuaternion?

MaLiN2223: If it was only about programming I would agree but unit quaternion and rotation quaternion are not semantically the same. Observe that sqrt(a) might be (in double precision) equal to 1 but a might be not.
Documentation about AlmostEqual method says "true if the two values differ by no more than 10 * 2^(-52); false otherwise." so we have 10*2^(-52) threshold.

Let's see the example :
a = 1 + 10\*2^(-51) is NOT almost equal to 1.0, 
sqrt(a) = sqrt(1 + 10\*2^(-51) ) < 1+10\*2^(-52) thus sqrt(a) is almost equal to 1.0

cdrnet: But does this different numerical criterion actually make sense? I.e. does it make sense that a specific quaternion can be at the same time considered as unit quaternion but not as rotation?

MaLiN2223: Well... I the code suggested me too much. You are correct. Rotation around non-unit quaternion doesn't make any sence because definion of the rotation uses unit quaternion. I will remove it in next commit ;)

JohanLarsson:
Suggestion: squash commits before merging, what do you think?

MaLiN2223:
I would love to but i can not handle "rebase -i". Would you be so kind and help me with that?

cdrnet:
No worries, I can merge it in whatever form it is :)

MaLiN2223:
I encountered on a problem - quaternion to power n (integer ) is not equal to n times multiplication of this quaternion.
Here is an example : 
Assume q = new Quaternion(1,1,1,1) therefore we have q\*q\*q = (-8,0,0,0)
but if q^3 ~= (-8, -1.28E-15i, -1.28E-15j, -1.28E-15k).

Unequality is caused by not 100% accuracy of pi - in upper example we calculate 
Sin(3 * ArcCos(1/2)) = Sin(pi) = 0 but in double calculation it is equal to about -3.22E-16.
Can we do anything about this? 

Function looks like this (not yet merged to branch) :
```
 public Quaternion Pow(double power)
           double arg = power * Arg;
            var u = Normalized;
            var cos = Math.Cos(arg);
            var sin = Math.Sin(arg);
            var output = Math.Pow(Norm, power) * (cos + u * sin);
```

This might cause very bad problems  - in example above q^3 should be scalar quaternion but it is not.

Only solutions I can see is to remove power from implementation or use some symbolic calculation to calculate those trig functions.

src/Spatial/Euclidean/Quaternion.cs
```diff
@@ -344,109 +342,81 @@ public Quaternion RotateUnitQuaternion(Quaternion unitQuaternion)
...
-            return new Quaternion(_w/d, _x/d, _y/d, _z/d);
+            return new Quaternion(-_w, -_x, -_y, -_z);
         }
-
         /// <summary>
-        /// Inverts this quaternion.
+        /// Inverts this quaternion. Inversing Zero returns Zero
         /// </summary>
-        public Quaternion Inverse()
+        public Quaternion Inversed
```
cdrnet: IMO this would better be a method as well (e.g. `Invert()`) instead of a property. Any thoughts?

src/Spatial/Euclidean/Quaternion.cs
```diff
@@ -166,43 +136,46 @@ public bool IsRotation
...
         {
-            return ToUnitQuaternion(0, _x, _y, _z);
+            get { return ToUnitQuaternion(0, _x, _y, _z); }
         }
 
         /// <summary>
         /// Returns a new normalized Quaternion q with the direction of this quaternion.
         /// </summary>
-        public Quaternion Sign()
+        public Quaternion Normalized
```
cdrnet: I'd tend to use a method for this as well (e.g. `Normalize()`), instead of a property. Thoughts?

JohanLarsson: Method would be consistent [with other parts of the lib](https://github.com/mathnet/mathnet-spatial/blob/master/src/Spatial/Euclidean/Vector3D.cs#L293-L296)

MaLiN2223: For Normailze and Invert i would use property because 
quaternion.Normalize() suggests (for me) that I am going to normalize quaternion not return a new one. Whereas property says "I am going to take quaternion and return normalized version of it" 
For me object method suggest that we are going to change this object. But yeah, for consistency I am going to change it.

JohanLarsson: I understand what you mean given all the mutable stuff in .net. Those are `public void Normalize()` though.
Still feel like a method is more right but can't word a reason why.
Would be nice if there was better language support for ignoring return values in C#. So you could say `yes I mean to ignore the return value here and get a compiler error if not. But that is another discussion.

cdrnet: Good points!

In general, at least for immutable types, the differentiation between when to use a method and when a property is about how expensive it is to get and if there is any non-trivial computation involved. If a user needs a value multiple times, when he sees a method he'll cache the result in a local variable, while a property he'll typically ask for it multiple times.

cdrnet:
Regarding `Pow` - could we maybe provide an overload (or `IntPow`) with integer-only exponents, which could be implemented in a different way?

src/Spatial/Euclidean/Quaternion.cs
MaLiN2223: You are both right. I am learning much from you, thanks.
I will change those properties to methods in next commit.

MaLiN2223:
I thought about it. I think it is a good idea to overload Pow. I think I can try to produce differen approach, in the worst case quaternion^n will be n times multiplication.

JohanLarsson:
@MaLiN2223 This is a very nice and much appreciated PR, hope you don't think the discussion is tiring.

MaLiN2223:
@JohanLarsson Not at all! I am just unexperienced developer thus it might seem that I surrender any disscusion but all I do is rethink my position.

MaLiN2223:
Unfortunately even [Chebyshev polynomials](https://en.wikipedia.org/wiki/Chebyshev_polynomials) are not good enough to make this work. Most probably creating normalized vector makes all the difference.
I have no other ideas, should I unaccurate or remove it?

My implementation was : 
```
public Quaternion Pow(int power)
{
    if (power == 0)
        return One;
    if (power == 1)
        return this;
    if (this == Zero || this == One)
                return this; 
    var pow = power / 2;
    var norm = Norm;
    var u = Normalize();
    var angle = Real / norm;
    var q = pow * 2 == power ? Math.Pow(NormSquared, pow) : Math.Pow(norm, power);
    return q * (ChybyshevCosPoli(power, angle) + u * ChybyshevSinPoli(power - 1, angle));
}
/// <summary>
/// Returns cos(n*arccos(x))
/// </summary> 
public static double ChybyshevCosPoli(int n, double x)
{ 
    if (n == 0)
        return 1.0;
    if (n == 1)
        return x;
    return 2 * ChybyshevCosPoli(n - 1, x) * x - ChybyshevCosPoli(n - 2, x);
}
/// <summary>
/// Returns sin(n*x)
/// </summary> 
public static double ChybyshevSinPoli(int n, double x)
{ 
    if (n == 0)
        return 1;
    if (n == 1)
        return 2 * x;
    return 2 * x * ChybyshevSinPoli(n - 1, x) - ChybyshevSinPoli(n - 2, x);
}
```

MaLiN2223:
I gave up with approximations. As far as I know there is no way to make int power accurate.
I think this PR is ready for merge.

cdrnet:
I've merged manually into master (due to the rebase going slightly too far). Thanks again!

JohanLarsson:
Thanks @MaLiN2223!

PR #29, opened 2016-02-19
Construct circle using 2 points and calculate circumference and area

EMostafaAli:
I added some functionality for circle that includes construct using 2 points and calculate diameter, circumference, and area.

I will try later to add a constructor to create the circle by 3 points

### DIFF
--- a/src/Spatial/Euclidean/Circle3D.cs
+++ b/src/Spatial/Euclidean/Circle3D.cs
@@ -15,5 +15,16 @@ namespace MathNet.Spatial.Euclidean
             this.Axis = axis;
             this.Radius = radius;
         }
+
+        public Circle3D(Point3D p1, Point3D p2, UnitVector3D axis)
+        {
+            this.CenterPoint = Point3D.MidPoint(p1, p2);
+            this.Axis = axis;
+            this.Radius = p1.DistanceTo(CenterPoint);
+        }
+
+        public double Diameter => 2 * Radius;
+        public double Circumference => 2 * Math.PI * Radius;
+        public double Area => Math.PI * Math.Pow(Radius, 2);
     }
 }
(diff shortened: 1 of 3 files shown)


JohanLarsson:
@cdrnet do you have an opinion on using c#6? I'm all for it.

cdrnet:
@JohanLarsson Sure, I'm fine with it in this project (but not in Numerics yet)

src/Spatial/Euclidean/Circle3D.cs
```diff
@@ -15,5 +15,16 @@ public Circle3D(Point3D centerPoint, UnitVector3D axis, double radius)
             this.Axis = axis;
             this.Radius = radius;
         }
+
+        public Circle3D(Point3D p1, Point3D p2, UnitVector3D axis)
+        {
+            this.CenterPoint = Point3D.MidPoint(p1, p2);
+            this.Axis = axis;
+            this.Radius = p1.DistanceTo(CenterPoint);
+        }
```
JohanLarsson: Hmm, maybe I was too quick at merging, don't really like this ctor as it is not very intuitive.

EMostafaAli:
I am an AutoCAD user, and it allows creating a circle using:

1- 2 points
2- 3 points
3- 2 tangents and a radius
4- 3 tangents


what is the problem with this construction?

cdrnet:
Looking at the code I can see that it creates essentially the smallest possible circle through two points, with the given axis (I'm sure there's a shorter/simpler way to describe this).

Maybe it would help to make this a static function instead, where the name implies what the two points are used for?

JohanLarsson:
Maybe just me not being used to it but it is a bit ambiguous. Could be centerpoint and point on tangent etc.

EMostafaAli:
For two points, there is infinite number of circles that can be drawn, on contrary two only 1 circle for 3 non-co-linear points).

Hence, this constructor, as in AutoCAD, draws the circle based on two endpoints of the diameter.

Please see the following link:
https://knowledge.autodesk.com/support/autocad-for-mac/learn-explore/caas/CloudHelp/cloudhelp/2015/ENU/AutoCAD-MAC-Core/files/GUID-C60B6D5D-AAEB-420F-917F-6E6B47E92F48-htm.html

PR #31, opened 2016-03-02
Issue19

mattj23:
ClosestPointTo methods added to Line3D and Line2D, the LineTo methods revised to utilize it, and the relevant unit tests.

Also an intersection method between Line2Ds and the unit tests.

### DIFF
--- a/src/Spatial/Euclidean/Line2D.cs
+++ b/src/Spatial/Euclidean/Line2D.cs
@@ -114,8 +114,12 @@ namespace MathNet.Spatial.Euclidean
             return this.StartPoint + alongVector;
         }
 
-
-        public Point2D IntersectWith(Line2D other)
+        /// <summary>
+        /// Compute the intersection between two lines
+        /// </summary>
+        /// <param name="other">The other line to compute the intersection with</param>
+        /// <returns>The point at the intersection of two lines, or null if the lines are parallelnu.</returns>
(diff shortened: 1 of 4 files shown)


JohanLarsson:
Thanks! Always nice with tests, the code looks good but I only skimmed it.

src/Spatial/Euclidean/Line2D.cs
```diff
@@ -99,8 +110,29 @@ public Line2D LineTo(Point2D p, bool mustStartBetweenAndEnd)
...
+        /// <returns>The point at the intersection of two lines, or null if the lines are parallelnu.</returns>
+        public Point2D? IntersectWith(Line2D other)
+        {
+            // http://stackoverflow.com/questions/563198/how-do-you-detect-where-two-line-segments-intersect
+            Point2D p = this.StartPoint;
+            Point2D q = other.StartPoint;
+            Vector2D r = this.StartPoint.VectorTo(this.EndPoint);
+            Vector2D s = other.StartPoint.VectorTo(other.EndPoint);
+
+            double t = (q - p).CrossProduct(s) / (r.CrossProduct(s));
```
JohanLarsson: `IsParallelTo(LIne2D other)` could perhaps be a separate method. Always tricky with tolerances for stuff like this.

JohanLarsson:
Thank you!

Issue #33, opened 2016-03-07
PolyLine2D/3D and Polygon2D classes


mattj23:
In my work I make common use of 2D and 3D polylines (a series of 2d or 3d points connected by line  segments) and 2D polygons (similar to a polyline except that the first and last points are connected, enclosing an area).  

I've written classes for my own use that handle these particular constructs and support basic operations, such as computation of length and area, projection of points onto them, convex hulls, etc, as well as unit tests for these operations.

Is this something that fits within the scope of Mathnet.Spatial?  If so, if I were to create a branch with these features in it would there be interest in reviewing it?

JohanLarsson:
Sounds awesome!

mattj23:
Take a look at this...the Polygon2D class is built on top of it, and the PolyLine3D class is essentially a 3D version of the same thing.

https://github.com/mattj23/mathnet-spatial/blob/issue33_polylines/src/Spatial/Euclidean/PolyLine2D.cs

JohanLarsson:
Can you make a PR with it? Nicer to review it that way.

PR #34, opened 2016-03-09
PolyLine2D class added

mattj23:
Basic PolyLine2D class submitted for review.  The Polygon2D class is built similarly to it, and the PolyLine3D class is basically a 3D version of the same thing.

Referencing Issue #33


src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+using System.ComponentModel;
+using System.Linq;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// The PolyLine2D class represents a 2D curve in space made up of line segments joined end-to-end, and is 
+    /// stored as a sequential list of 2D points.
+    /// </summary>
+    public class PolyLine2D : IEnumerable<Point2D>
```
JohanLarsson: IReadonlyList<Point2D> makes sense here unless bumping framework version is an issue cc/ @cdrnet

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+    {
+        // Internal storage for the points
+        private List<Point2D> _points;
+
+        /// <summary>
+        /// Returns the number of points in the polyline
+        /// </summary>
+        public int Count => this._points.Count;
```
JohanLarsson: NIT: we don't use `this.`in the rest of the code.

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        /// <summary>
+        /// Returns the length of the polyline as the sum of the length of the individual segments
+        /// </summary>
+        public double Length => this.GetPolyLineLength();
+
+        /// <summary>
+        /// Default constructor, creates an empty PolyLine2D
+        /// </summary>
+        public PolyLine2D() : this(Enumerable.Empty<Point2D>())
```
JohanLarsson: Is this ever useful since there is no way to add anything to it later?
`public static readonly PolyLine2D Empty = new PolyLine2D(Enumerable.Empty<Point2D>())`
Is an alternative

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        /// <summary>
+        /// Access a point in the polyline by index number
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        public Point2D this[int key]
+        {
+            get { return this._points[key]; }
+            set { this._points[key] = value; }
```
JohanLarsson: Not sure about mutable, not saying it is wrong just not sure.

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+        /// <summary>
+        /// Get the point at a specified distance along the curve.  A negative argument will return the first point,
+        /// an argument greater than the length of the curve will return the last point.
+        /// </summary>
+        /// <param name="lengthFromStart">The distance from the first point along the curve at which to return a point</param>
+        /// <returns></returns>
+        public Point2D GetPointAtLengthFromStart(double lengthFromStart)
+        {
+            double length = this.Length;
+            if (lengthFromStart >= length)
```
JohanLarsson: Not sure what is best here, throwing? Returning null? As written?

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        /// <summary>
+        /// Get the point at a fractional distance along the curve.  For instance, fraction=0.5 will return
+        /// the point halfway along the length of the polyline.
+        /// </summary>
+        /// <param name="fraction">The fractional length at which to compute the point</param>
+        /// <returns></returns>
+        public Point2D GetPointAtFractionAlongCurve(double fraction)
+        {
+            return this.GetPointAtLengthFromStart(fraction * this.Length);
```
JohanLarsson: Fraction must be 0 < fraction < 1 here right? Check and throw argument exception feels right if so.

mattj23:
I think when I originally wrote it I intended for there to be a way to add and remove points, then didn't end up needing it for the project I was working on.

Do you have any particular preference?  I can think of a few situations in which it would be favorable to modify the contents of the list rather than creating a whole new object, specifically in cases where these polylines contain thousands or tens of thousands of points, but I'm not committed either way.

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+        /// <summary>
+        /// Reduce the complexity of a manifold of points represented as an IEnumerable of Point2D objects.
+        /// This algorithm goes through each point in the manifold and computes the error that would be introduced
+        /// from the original if that point were removed.  Then it removes nonadjacent points to produce a 
+        /// reduced size manifold.
+        /// </summary>
+        /// <param name="points"></param>
+        /// <param name="tolerance"></param>
+        /// <returns></returns>
+        private static IEnumerable<Point2D> ReduceComplexitySingleStep(IEnumerable<Point2D> points, double tolerance)
```
JohanLarsson: NIT: we have followed StyleCop ordering public before private, static before instance etc.

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        /// <summary>
+        /// Reduce the complexity of a manifold of points represented as an IEnumerable of Point2D objects by
+        /// iteratively removing all nonadjacent points which would each result in an error of less than the
+        /// single step tolerance if removed.  Iterate until no further changes are made.
+        /// </summary>
+        /// <param name="points"></param>
+        /// <param name="singleStepTolerance"></param>
+        /// <returns></returns>
+        public static PolyLine2D ReduceComplexity(IEnumerable<Point2D> points, double singleStepTolerance)
```
JohanLarsson: I'll need to read this more later but all these `.ToList()` looks expensive.

JohanLarsson:
My preference is readonly/immutable for all the things :) As it closes the door to so many bugs.

mattj23:
Updated with a new commit that has the Polygon2D and PolyLine3D classes.  There's no unit tests for PolyLine3D, but Polygon2D has 80% coverage and PolyLine2D has 93% coverage, by lines.

src/Spatial/Euclidean/PolyLine2D.cs
mattj23: The ReduceComplexity operation is definitely expensive in general, it's taking a raw polyline and removing redundant points. It's typically only ever done once on a set of points.  If there's a way to rewrite it faster I'm open to suggestions.

JohanLarsson:
I would merge this if it was immutable. Have you seen the [chat](https://gitter.im/mathnet/mathnet-spatial) @mattj23?

mattj23:
Alright, I will make these classes immutable structs.  I may need some guidance as I go forward.

JohanLarsson:
@cdrnet what do you think about implementing polyline as a struct? Feels like a class to me.

mattj23:
What is the "right" way to make them immutable?  I really have no idea what I'm doing here.

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -0,0 +1,219 @@
...
+using System.ComponentModel;
+using System.Linq;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// The PolyLine2D class represents a 2D curve in space made up of line segments joined end-to-end, and is 
+    /// stored as a sequential list of 2D points.
+    /// </summary>
+    public struct PolyLine2D : IEnumerable<Point2D>
```
JohanLarsson: Use IReadOnlyList<Point2D> here or is it a problem that it is .net 4 @cdrnet?

JohanLarsson:
@mattj23 By removing all mutating members. The only one I spotted is the indexer, remove `set { this._points[key] = value; }`

JohanLarsson:
If you feel like it you can rebase on master.

JohanLarsson:
@mattj23 Looks like one test is failing.

mattj23:
Thanks, I fixed it.  There's one more piece of functionality I want to add.

mattj23:
Actually, scratch that, I'll do it separately, since it's more of a feature for Line3Ds that would then get used by PolyLine3D.

JohanLarsson:
Sorry about the delays and the slow review.

Issue #32, opened 2016-03-07
IsParallelTo method for Line2D (and Line3D)


mattj23:
Per discussion in the issue19 thread, is this what you're thinking?

```csharp

        /// <summary>
        /// Compute the intersection between two lines
        /// </summary>
        /// <param name="other">The other line to compute the intersection with</param>
        /// <returns>The point at the intersection of two lines, or null if the lines are parallel.</returns>
        public Point2D? IntersectWith(Line2D other)
        {
            if (this.IsParallelTo(other))
                return null;

            // http://stackoverflow.com/questions/563198/how-do-you-detect-where-two-line-segments-intersect
            Point2D p = this.StartPoint;
            Point2D q = other.StartPoint;
            Vector2D r = this.StartPoint.VectorTo(this.EndPoint);
            Vector2D s = other.StartPoint.VectorTo(other.EndPoint);

            double t = (q - p).CrossProduct(s) / (r.CrossProduct(s));

            return p + t * r;
        }

        /// <summary>
        /// Checks to determine whether or not two lines are parallel to each other, using the cross product
        /// </summary>
        /// <param name="other">The other line to check this one against</param>
        /// <returns>True if the lines are parallel, false if they are not</returns>
        public bool IsParallelTo(Line2D other)
        {
            return Math.Abs(this.Direction.CrossProduct(other.Direction)) < double.Epsilon;
        }

```

I'll happily update it and write some unit tests.

JohanLarsson:
Yeah, I'm thinking it can be good for two reasons:
1) Useful method in itself.
2) Improved readability where used internally.

The tricky part is tolerance, one idea could be to pass in an `Angle` as tolerance. What do you think?
I'd expect double.Epsilon to be too small for practical use.

mattj23:
How about this:

```csharp

        /// <summary>
        /// Compute the intersection between two lines
        /// </summary>
        /// <param name="other">The other line to compute the intersection with</param>
        /// <returns>The point at the intersection of two lines, or null if the lines are parallel.</returns>
        public Point2D? IntersectWith(Line2D other)
        {
            if (this.IsParallelTo(other))
                return null;

            // http://stackoverflow.com/questions/563198/how-do-you-detect-where-two-line-segments-intersect
            Point2D p = this.StartPoint;
            Point2D q = other.StartPoint;
            Vector2D r = this.StartPoint.VectorTo(this.EndPoint);
[...]

PR #37, opened 2016-03-10
Issue32

mattj23:
Pull request for Issue #32 submitted for review.

This one turned into something big.  Changed default tolerances to 1e-10, since 1e-6 (which was on the Vector3D class's IsParallelTo method) ended up producing incorrect results with vectors different in the third decimal place.

Lots of unit tests, should be 100% coverage of the methods I added and the old IsParallelTo methods.  

I did not remove all of the old tests, but maybe they ought to be.

### DIFF
--- a/src/Spatial/Euclidean/Line2D.cs
+++ b/src/Spatial/Euclidean/Line2D.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using MathNet.Spatial.Units;
 
 namespace MathNet.Spatial.Euclidean
 {
@@ -118,9 +119,12 @@ namespace MathNet.Spatial.Euclidean
         /// Compute the intersection between two lines
         /// </summary>
         /// <param name="other">The other line to compute the intersection with</param>
-        /// <returns>The point at the intersection of two lines, or null if the lines are parallelnu.</returns>
+        /// <returns>The point at the intersection of two lines, or null if the lines are parallel.</returns>
         public Point2D? IntersectWith(Line2D other)
         {
+            if (this.IsParallelTo(other))
+                return null;
+
             // http://stackoverflow.com/questions/563198/how-do-you-detect-where-two-line-segments-intersect
             Point2D p = this.StartPoint;
             Point2D q = other.StartPoint;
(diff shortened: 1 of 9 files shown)


JohanLarsson:
Thanks!

src/Spatial/Euclidean/Line2D.cs
```diff
@@ -129,12 +135,53 @@ public Point2D ClosestPointTo(Point2D p, bool mustBeOnSegment)
...
 
+        /// <summary>
+        /// Checks to determine whether or not two lines are parallel to each other, using the dot product within 
+        /// the double precision specified in the MathNet.Numerics package.
+        /// </summary>
+        /// <param name="other">The other line to check this one against</param>
+        /// <returns>True if the lines are parallel, false if they are not</returns>
+        public bool IsParallelTo(Line2D other)
+        {
+            return this.Direction.IsParallelTo(other.Direction, Precision.DoublePrecision * 2);
```
JohanLarsson: What was the reason for the *2 again?

mattj23: The test case failed with two identical vectors when it was just Precision.DoublePrecision.  In fact, the mismatch was exactly DoublePrecision * 2, so the comparison had to become a <= in order for it to return correct results.

src/SpatialUnitTests/Euclidean/Line2DTests.cs
```diff
@@ -174,8 +175,35 @@ public void IntersectWithTest(string s1, string e1, string s2, string e2, string
...
+        [TestCase("0,0", "0.00001,-1.0000", "1,1", "1,2", false)]
+        public void IsParallelToWithinDoubleTol(string s1, string e1, string s2, string e2, bool expected)
+        {
+            var line1 = Line2D.Parse(s1, e1);
+            var line2 = Line2D.Parse(s2, e2);
+            
+            Assert.AreEqual(expected, line1.IsParallelTo(line2));
+        }
+
+        [TestCase("0,0", "0,1", "1,1", "1,2", 0.01, true)]
```
JohanLarsson: `[TestCase("0,0", "0,1", "0,1", "0,0", 0.01, true|false)]`
Here we a have a decision to make.

mattj23: A friend of mine is a professor of mathematics, I'll reach out to him and see what he thinks.  If he can't answer the question I suspect he'll know someone who can.

JohanLarsson:
@mattj23 sorry about the delay, really busy here.

PR #42, opened 2016-05-22
Line3d intersections

mattj23:
This code performs searching for closest distances between two Line3D objects, analogous to computing intersections between two-dimensional lines.

The difference is that with lines in 3d, we often want to know the location of the closest distance between the two lines even if they do not mathematically intersect.  In such a case the closest span is best represented by a pair of points, one which lies on the first line, and the other which lies on the second line.  These two points are identical when the lines actually intersect, or are separated by a finite distance if they are skew.

The best way I could think of to represent this point pair was with a Tuple<Point3D, Point3D>, since a Line3D will throw an error if the start and end points are identical.


src/Spatial/Euclidean/Line3D.cs
```diff
@@ -181,6 +182,91 @@ public bool IsParallelTo(Line3D other, Angle angleTolerance)
         }
 
         /// <summary>
+        /// Computes the pair of points which represent the closest distance between this Line3D and another Line3D, with the first
+        /// point being the point on this Line3D, and the second point being the corresponding point on the other Line3D.  If the lines
+        /// intersect the points will be identical, if the lines are parallel the first point will be the start point of this line.
+        /// </summary>
+        /// <param name="other">line to compute the closest points with</param>
+        /// <returns>A tuple of two points representing the endpoints of the shortest distance between the two lines</returns>
+        public Tuple<Point3D, Point3D> ClosestPointsBetween(Line3D other)
```
JohanLarsson: What about returning a line?

src/Spatial/Euclidean/Line3D.cs
```diff
@@ -181,6 +182,91 @@ public bool IsParallelTo(Line3D other, Angle angleTolerance)
...
+
+            var fourChecks = new Tuple<double, Tuple<Point3D, Point3D>>[]
+            {
+                Tuple.Create(r1.DistanceTo(this.StartPoint), Tuple.Create(r1, this.StartPoint)),
+                Tuple.Create(r2.DistanceTo(this.EndPoint), Tuple.Create(r2, this.EndPoint)),
+                Tuple.Create(r3.DistanceTo(other.StartPoint), Tuple.Create(r3, other.StartPoint)),
+                Tuple.Create(r4.DistanceTo(other.EndPoint), Tuple.Create(r4, other.EndPoint))
+            };
+
+            return fourChecks.Aggregate((i1, i2) => i1.Item1 < i2.Item1 ? i1 : i2).Item2;
```
JohanLarsson: Allocating an array and tuples and linq like this is pretty expensive. Probably not a problem but can be a future refactoring opportunity for perf.

src/Spatial/Euclidean/Point3D.cs
```diff
@@ -226,6 +226,16 @@ public static Point3D IntersectionOf(Plane plane, Ray3D ray)
             return plane.IntersectionWith(ray);
         }
 
+        public static Point3D operator +(Vector3D v, Point3D p)
```
JohanLarsson: For some reason I did not add these, point + vector felt natural, vector + point did not (for me).
Maybe this can be a separate PR to give others the opportunity to have a look?

JohanLarsson:
Thanks! 
Looks nice, nice code, nice comments and nice tests as always. 
Only question I have is the name of the method and the return type.

src/Spatial/Euclidean/Line3D.cs
cdrnet: To me the chosen return type and naming does make sense. This is really about two points (which can be the same), not technically a line.

There could be another routine on top which does returns a line, but this other routine would then have to fail in some cases (e.g. when they intersect).

src/Spatial/Euclidean/Point3D.cs
mattj23: My thought here is that addition is commutative, and so to have the order matter diverges from the math itself. I will separate this out so that it doesn't hold up the PR.

src/Spatial/Euclidean/Line3D.cs
mattj23: Let me change it to a loop. It was late and I was in a Python mindset, it doesn't need to be so complicated.

PR #43, opened 2016-05-23
Added operators to allow Point3D + Vector3D to be commutative

mattj23:
Points are technically just vectors from the origin, and vector addition is commutative, so it makes sense that adding a point to a vector should be the same as adding a vector to a point.  This adds the operators to allow a Vector3D to be added to a Point3D, which will hopefully help programmers using the library when translating formulas and equations from the math world to computer code, preventing them from having to switch the order of the operands when they don't match with what the MathNet libraries allow.

### DIFF
--- a/src/Spatial/Euclidean/Point3D.cs
+++ b/src/Spatial/Euclidean/Point3D.cs
@@ -226,6 +226,16 @@ namespace MathNet.Spatial.Euclidean
             return plane.IntersectionWith(ray);
         }
 
+        public static Point3D operator +(Vector3D v, Point3D p)
+        {
+            return p + v;
+        }
+
+        public static Point3D operator +(UnitVector3D v, Point3D p)
+        {
+            return p + v;
+        }
+
         public static Point3D operator +(Point3D p, Vector3D v)
         {
             return new Point3D(p.X + v.X, p.Y + v.Y, p.Z + v.Z);


AQuentin2033:
adding a vector to a point makes sense, not convinced that adding a point to a vector does.

mattj23:
@AQuentin2033 Can you clarify what you mean when you say it doesn't make sense?

A point is represented in Euclidian n-space by a vector from the origin, thus the entire reason it is possible to add a vector to a point is because both are vectors to begin with.  As I mentioned before, vector addition is commutative, so it breaks the mathematical formalism to consider it valid to add a vector to a point but not a point to a vector.  Either a point is a vector and they can be added together in any order, or a point isn't a vector and we should remove the ability to add them at all and replace the addition operator with a "shift" or "displace" method.

AQuentin2033:
In mathematics there is a definition describing the adding of a vector to a point.  There is no corresponding definition for adding a point to anything.  I realize what you are saying about their technical equivalence, and that you could substitute any point for a vector.  In mathematics terms, I see that more of an explicit operation than an implicit one.  So you would do something like `v + p.ToVector()` or eventually if we wanted to implement explicit conversion `v + (Vector2D)p.`  but to do `v + p` would involve an implicit conversion which may lead to ambiguity and confusion.  Particularly if you couldn't use a point everywhere a vector was defined.  The other way around `p + v ` is a formally defined operation in mathematics and thus holds no such confusion.

So in my view, conceptually [points are not vectors](http://graphics.cs.wisc.edu/WP/cs559-fall2014/2014/08/28/points-vectors-and-coordinate-systems-why-are-points-and-vectors-different/) but we still have add because it is a formal operation defined in mathematics on a vector.
Make sense?

mattj23:
I see where you're coming from, but I feel that it's an over-interpretation of what the conceptual difference means for a a point.

Sure, a point is a different thing conceptually than a vector.  A point is a set element.  A point in n-space can be embedded in a higher or lower dimensional manifold.  A point can represent a feature, or be part of some other topological or metric space.  All of these things encompass more than a simple vector.

_However_, when a point in R^n space is represented as an ordered sequence of numbers which define its offset from the origin, that _representation is absolutely, 100% mathematically a vector_.

So what I'm struggling with is this: does the Point3D class encapsulate the general concept of a point, which is conceptually different from a vector, or does it encapsulate the 3-space representation of a point as an offset from the origin, which _is actually a vector_?  The implementation of Point3D suggests the latter.  The inclusion of vector addition suggests the latter as well.

If Point3D encapsulates the concept of a point, then any addition with a vector involves a cast from the point to its offset-representation and there's no order of operands which will sanitize that.  If this is the case it would make the most mathematical sense to remove addition as a fundamental operator and replace it with something explicit like p.DisplaceBy(v).

If Point3D encapsulates the offset representation then addition with a vector is fine, and in vector addition the order of the operands is unimportant.

What we have now is a class that is ambiguously between the two, and in doing so diverges from the rules of either plausible mathematical interpretation.  I'm not a project maintainer so it's not my decision to make, but as a user and an occasional contributor it's frustrating to have these abstractions which diverge from the formalisms they're intended to abstract (much like the Line*D classes, whose implementations represent line segments rather than lines).

(Regardless, I appreciate this project and thanks for helping maintain great open source software)

RiSearcher:
Even if point is represented as offset from origin, point is not equivalent to vector. When you transform point and vector from one coordinate system to another, you use different sets of rules.

AQuentin2033:
Interesting, I would have said that Point3D should encapsulate the general concept of a point.  The whole piece about general transforms, translations,, scaling, rotations and the like for me all need to be redressed in this library and I think if the corresponding methods existed so that there was a common way of translating any geometric item in space from one position to another then this probably wouldn't be a discussion.
So in that light I would be more in favor of adding TranslateBy(v) and obsoleting the additive methods on Point2D/3D than making point and vector equivalent

re Line2D/3D I 100% agree- btw, we have already created LineSegment2D/3D and removed internal library use of Line2D/3D.  When the next release is made (just waiting on Mathnet.Numerics 4.0 to be released) we will obsolete the existing Line2D/3D types and the release after remove them.

Issue #50, opened 2016-12-13
Ordering of Polygon2D.GetConvexHullFromPoints uses AngleTo instead of SingedAngleTo


kebabbi:
This results in a wrong ordering of the points on the convex hull.

JohanLarsson:
Assuming this was fixed.

PR #51, opened 2016-12-13
Order points on the convex hull clockwise

kebabbi:
Solved issue #50

### DIFF
--- a/src/Spatial/Euclidean/Polygon2D.cs
+++ b/src/Spatial/Euclidean/Polygon2D.cs
@@ -83,8 +83,12 @@ namespace MathNet.Spatial.Euclidean
         /// two dimensional convex hull, returning it as a Polygon2D object.  
         /// </summary>
         /// <param name="pointList"></param>
+        /// <param name="clockWise">
+        /// In which direction to return the points on the convex hull.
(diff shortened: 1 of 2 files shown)


src/Spatial/Euclidean/Polygon2D.cs
```diff
@@ -133,7 +133,7 @@ public static Polygon2D GetConvexHullFromPoints(IEnumerable<Point2D> pointList)
             // Order the hull points by angle to the centroid
             Point2D centroid = Point2D.Centroid(hullPoints);
             Vector2D xAxis = new Vector2D(1, 0);
-            var results = (from x in hullPoints select new Tuple<Angle, Point2D>(centroid.VectorTo(x).AngleTo(xAxis), x)).ToList();
+            var results = (from x in hullPoints select new Tuple<Angle, Point2D>(centroid.VectorTo(x).SignedAngleTo(xAxis, true), x)).ToList();
```
JohanLarsson: This is going to be pretty slow. Also no tests?

kebabbi: Could you explain why this is going to be slow? In any case sorting on absolute angle does not seem really useful to me. Maybe I could make this sorting optional?

Issue #47, opened 2016-10-17
Consider offering XML serialization as separate product


JKSnd:
The project I'm using mathnet spatial for uses json for serialization instead of xml.
It isn't ideal that my project takes on an xml dependency that is never used.

Can serialization be broken out into a separate nuget package?

JohanLarsson:
How do you suggest doing this given how `XmlSerializer` works?
I agree that the dependency on `Sytem.Xml` is not ideal.

JKSnd:
I would move the responsibility of XML serialization into a set of data transfer objects ([DTOs](https://en.wikipedia.org/wiki/Data_transfer_object)) and use a mapping layer to convert between the core types and the DTOs. This pattern could then be used to add JSON serialization similarly.

Then a developer could add serialization support to their project through Nuget by installing MathNet.Spatial.Serialization.Xml, for example. [Units.Net](https://github.com/anjdreas/UnitsNet) does something similar.

AQuentin2033:
I can move support for serialization from the system.runtime.serialization (DataContracts etc) into a separate (.netstandard2.0 - possibly 1.6) assembly (and at the same time improve serialization support).  This is fairly straight forward and works well and as you say should extend to being able to support json too.

However, with regards to the native XmlSerializer even with DTOs the use experience is not great as you cannot (at least as far as I can find) intercept the type lookup for nested types.  This would mean converting a type tree into a surrogate type tree before serializing and back again after deserialization.  This would only work if you were serializing the objects as standalone objects but would fail in the case you embedded a geometry type into your own serializable class.  Not really an ideal experience.
An option would be to drop support for XmlSerializer and either just support data contracts or also support a 3rd party xml serializing solution like [ExtendedXmlSerializer](https://www.nuget.org/packages/ExtendedXmlSerializer/#) in order to be able to keep existing format compatibility for people who have saved representations of the current types.

Another thing to keep in mind is that the current xml serialization support is incomplete with not all types in the library supporting serialization, so either way there is something to be addressed here.

JohanLarsson:
How big problem is the dependency on `System.Xml`?
I agree it is not nice but can be convenient.
Next question is how big is the upside?, I don't know if `XmlSerializer`is used much.

@AQuentin2033 Agreed we should have all types implement `IXmlSerializable` if we have the dependency.

JohanLarsson:
[...]

PR #60, opened 2017-11-29
Serialize

AQuentin2033:
As requested in issue #47 this change introduces a separate assembly for Xml Serialization.

As currently implemented only system.runtime.serialization (DataContracts) would be supported.  This would allow the subsequent removal of all Xml references in the main project.

Optionally I could also add in support for the 3rd party ExtendedXmlSerializer to be able to support existing XmlSerialization without retaining the reference to xml in the main project.

For the moment the test cases are in a separate library - perhaps it is better of they are in the common test library.


cdrnet:
Is this more about Xml or about DataContracts? Note that the latter is not bound to Xml, but can also be used with Protocol Buffers, Json, FsPickler and others. Just wondering about the naming, not about the separation as is.

AQuentin2033:
Well it started with the request to split out the Xml hence the name, but you are indeed correct in that it is potentially broader.  It implements ISerializationSurrogateProvider which is the .net standard replacement for IDataContractSurrogate .  As far as I am aware so far the only serializer which has been updated so far with support for this interface is the DataContractSerializer.  Although I imagine that the others framework classes will gain it over time.  
Protobuf-net has its own method for handling surrogates but it looks straightforward and I could add that support for that too if its interesting to anyone.
I currently do have the code build the SurrogateSelector for a older Soap/BinaryFormatters (although this could admittedly do with a few more test cases to be sure its ok)

Would simply Mathnet.Spatial.Serialization be a better name?

AQuentin2033:
I have re worked this in another branch, so this is obsolete

Issue #30, opened 2016-02-25
Geometrical figures


ForNeVeR:
Hello. Is Math.NET Spatial is the proper place to add some geometrical figures such as cubes, spheres, cylinders etc?

I am working on multiple scientific projects and really tired of copying the definitions for all the basic geometry back and forth. Math.NET Spatial have all these great definitions for points and vectors, and that's cool already. But could we have the figures also?

I'm looking forward to contribute some stuff, but I am still in doubts about 
a) whether Math.NET Spatial is the proper place to put this stuff
b) about the quality of my code - I am not the professional mathematician at all, I just have to work with some simple spatial definitions on a day to day basis.

What can you guys say? Maybe some ideas about what absolute minimal methods/definitions we should add to the figures or how the code should be structured (with inheritance from base Figure class or some another functional stuff)?

I am thinking about adding some basic stuff first (such as `Box3D` and `Sphere3D` maybe?) and adding the new methods (intersections/etc) based on the community feedback.

Maybe someone is already working on it?

cdrnet:
I'm not aware anyone us currently working on this already, but figures like cubes and cones would indeed fit perfectly into the scope of this project.

gsomix:
Good suggestions. I'm about to assist.

gsomix:
> what absolute minimal methods/definitions we should add to the figures

@ForNeVeR I think we can define most of basic figures using linear dimensions (radius of sphere, height and radius of cylinder, etc) and local coordinate system. Minimal useful set of methods may include points, vertices and faces sampling.

JohanLarsson:
@gsomix about `CoordinateSystem`: there is need for big cleanup and API improvement.
- Uses `Matrix` internally, probably suboptimal for performance.
- The public API is borrowed from Autodesk Inventor, not happy with it at all.

The reason I'm writing this is so you guys don't hesitate to touch it if you find things that look dumb. I sometimes regret including it.
@ForNeVeR sounds great!

ForNeVeR:
Well, guys, I'm already prototyping something and hopefully will be able to show you the API design in the nearest few days (nothing big though - just a simple API with a few box-based operations). I'll create a work-in-progress PR and we'll be able to discuss it in a more details.

PR #61, opened 2017-12-01
Circle and StraightLine

AQuentin2033:
Starting to address issue #30.  First 2 objects Circle and StraightLine.

I also took the opportunity to clean up some of the missing comments and test errors (duplicated tests ignored by nunit)

As an open question for the future - would there be any objection to adding a reference to Mathnet.Symbolics ?  I would like to add some equation support to the geometric shapes (particularly lines)


src/Spatial/Euclidean/Circle.cs
```diff
@@ -0,0 +1,147 @@
...
+{
+    /// <summary>
+    /// Describes a standard 2 dimentional circle
+    /// </summary>
+    public class Circle : IEquatable<Circle>
+    {
+        /// <summary>
+        /// The center point of the circle
+        /// </summary>
+        public Point2D Center { get; }
```
JohanLarsson: Minor style nag: we use style cop conventions meaning ctor before properties. Not very important but makes future diffs cleaner.

src/Spatial/Euclidean/Circle.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// Describes a standard 2 dimentional circle
+    /// </summary>
+    public class Circle : IEquatable<Circle>
```
JohanLarsson: We assume namespances here, more consistent with the rest of the lib would be Circle2D.

src/Spatial/Euclidean/ILine.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="y">An optional y coordinate</param>
+        /// <returns>A list of crossing points</returns>
+        IEnumerable<Point2D> YIntercept(double y = 0);
+
+        /// <summary>
+        /// Determines if a given point is on the line
+        /// </summary>
+        /// <param name="point">The point to be tested</param>
+        /// <returns>True if the point is on the line</returns>
+        bool IsOnLine(Point2D point);
```
JohanLarsson: `line.IsOnLine(p)` will read a little weird. Not sure if `IsThrough` is good. An overload with a tolerance might makes sense.

src/Spatial/Euclidean/Point2D.cs
```diff
@@ -54,36 +73,65 @@ public Point2D(double[] data)
...
+        /// <summary>
+        /// Returns the centeroid or center of mass of any set of points
+        /// </summary>
+        /// <param name="points">a list of points</param>
+        /// <returns>the centeroid point</returns>
         public static Point2D Centroid(params Point2D[] points)
         {
             return new Point2D(
                 points.Average(point => point.X),
                 points.Average(point => point.Y));
```
JohanLarsson: I should fix this to not use linq, very inefficient implementation here.

src/Spatial/Euclidean/StraightLine.cs
```diff
@@ -0,0 +1,258 @@
...
+
+        /// <summary>
+        /// Returns the hashocde for a Straightline
+        /// </summary>
+        /// <returns>A hashcode</returns>
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (Gradient.GetHashCode() * 397) ^ yoffset.GetHashCode();
```
JohanLarsson: Why is xoffset not used here? Also yoffset is not `readonly`

src/Spatial/Euclidean/StraightLine.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <param name="p2">A point on the line</param>
+        public StraightLine(Point2D p1, Point2D p2)
+        {
+            if (p1 == p2)
+                throw new ArgumentException("A line cannot be formed from two identical points");
+            Gradient = (p2.Y - p1.Y) / (p2.X - p1.X);
+            setOffset(p1.X, p1.Y);
+        }
+
+        private void setOffset(double x, double y)
```
JohanLarsson: I prefer `private static void SetOffset(ref double x, ref double y)`. That allows the fields to be `readonly`.
Rename to `SetOffset`, also private methods last according to stylecop.

src/Spatial/Euclidean/Circle.cs
AQuentin2033: ok, will update styles.  I didn't use readonly public fields deliberately in order to be able to later add support for lazy calculation where multiple construction methods exist

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -0,0 +1,154 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// Describes a standard 2 dimentional circle
+    /// </summary>
+    public class Circle2D : IEquatable<Circle2D>
```
JohanLarsson: Make this a struct.

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -0,0 +1,154 @@
...
+            Radius = radius;
+        }
+
+        /// <summary>
+        /// Creates a Circle from three points which are not colinnear
+        /// </summary>
+        /// <param name="pointA">A point on the circle</param>
+        /// <param name="pointB">A point on the circle</param>
+        /// /// <param name="pointC">A point on the circle</param>
+        public Circle2D(Point2D pointA, Point2D pointB, Point2D pointC)
```
JohanLarsson: This is perhaps nicer as a factory method: `var circle = Cricle2D.FromThreePoints(p1, p2, p3)` Not sure if `FromThreePoints` is a good name, a bit redundant with three give three args.

src/Spatial/Euclidean/Line2D.cs
```diff
@@ -252,4 +252,4 @@ public override int GetHashCode()
         }
         # endregion
     }
-}
\ No newline at end of file
+}
```
JohanLarsson: Remove this from the diff.

src/Spatial/Euclidean/PlaneGeometry/LineSegment.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean.PlaneGeometry
+{
+    public class LineSegment : LineSegment<StraightLine>
```
JohanLarsson: class and should probably be named LineSegment2D

AQuentin2033:
split as requested

PR #67, opened 2017-12-02
StraightLine2D Struct

AQuentin2033:
This Struct represents a geometric line of the equation y=mx + b

This patch also introduces ILine interface for all types of (future) geometric lines

### DIFF
--- a/src/Spatial/Euclidean/ILine.cs
+++ b/src/Spatial/Euclidean/ILine.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// This interface represents a line in cartesian space
+    /// </summary>
+    public interface ILine : IEquatable<ILine>
+    {
+        /// <summary>
+        /// Returns the points at which the line crosses on the x axis
+        /// </summary>
+        /// <param name="x">An optional x coordinate</param>
+        /// <returns>A list of crossing points</returns>
+        IEnumerable<Point2D> XIntercept(double x = 0);
+
+        /// <summary>
+        /// Returns the points at which the line crosses on the y axis.
+        /// </summary>
+        /// <param name="y">An optional y coordinate</param>
+        /// <returns>A list of crossing points</returns>
+        IEnumerable<Point2D> YIntercept(double y = 0);
+
+        /// <summary>
+        /// Determines if a given point is on the line
+        /// </summary>
+        /// <param name="point">The point to be tested</param>
+        /// <returns>True if the point is on the line</returns>
+        bool IsThrough(Point2D point);
+
+        /// <summary>
+        /// Determines if a given point is on the line
+        /// </summary>
+        /// <param name="x">The x coordinate of the point</param>
+        /// <param name="y">The y coordinate of the point</param>
+        /// <returns>True if the point is on the line</returns>
+        bool IsThrough(double x, double y);
+
+        /// <summary>
+        /// Determines if a line is parallel to another line
+        /// </summary>
+        /// <param name="line">the line to be compared against</param>
+        /// <returns>True if the lines are parallel</returns>
+        bool IsParallel(ILine line);
+    }
+}
--- a/src/Spatial/Euclidean/ILine2D.cs
+++ b/src/Spatial/Euclidean/ILine2D.cs
@@ -8,21 +8,21 @@ namespace MathNet.Spatial.Euclidean
     /// <summary>
     /// This interface represents a line in cartesian space
     /// </summary>
-    public interface ILine : IEquatable<ILine>
+    public interface ILine2D : IEquatable<ILine2D>
     {
         /// <summary>
         /// Returns the points at which the line crosses on the x axis
         /// </summary>
         /// <param name="x">An optional x coordinate</param>
         /// <returns>A list of crossing points</returns>
-        IEnumerable<Point2D> XIntercept(double x = 0);
+        Point2D[] XIntercept(double x = 0);
 
         /// <summary>
         /// Returns the points at which the line crosses on the y axis.
         /// </summary>
         /// <param name="y">An optional y coordinate</param>
         /// <returns>A list of crossing points</returns>
-        IEnumerable<Point2D> YIntercept(double y = 0);
+        Point2D[] YIntercept(double y = 0);
 
         /// <summary>
         /// Determines if a given point is on the line
@@ -44,6 +44,6 @@ namespace MathNet.Spatial.Euclidean
         /// </summary>
         /// <param name="line">the line to be compared against</param>
         /// <returns>True if the lines are parallel</returns>
-        bool IsParallel(ILine line);
+        bool IsParallel<T>(T line) where T : ILine2D;
     }
 }
--- a/src/Spatial/Euclidean/StraightLine2D.cs
+++ b/src/Spatial/Euclidean/StraightLine2D.cs
@@ -141,7 +141,12 @@ namespace MathNet.Spatial.Euclidean
         /// <returns>A point midway between the two provided points</returns>
         public Point2D MidPoint(Point2D pointA, Point2D pointB)
         {
-            return new Point2D((pointA.X + pointB.X) / 2, (pointA.Y + pointB.Y) / 2);
+            if (IsThrough(pointA) && IsThrough(pointB))
+            {
+                return new Point2D((pointA.X + pointB.X) / 2, (pointA.Y + pointB.Y) / 2);
+            }
+
+            throw new ArgumentException("Provided points are not along the line");
         }
 
         /// <summary>
@@ -212,7 +217,14 @@ namespace MathNet.Spatial.Euclidean
         /// <returns>True if the point is on the line</returns>
         public bool IsThrough(double x, double y)
         {
-            return y == (Gradient * x) + yoffset;
+            if (double.IsInfinity(Gradient))
+            {
+                return x == xoffset;
+            }
+            else
+            {
+                return y == (Gradient * x) + yoffset;
+            }
         }
 
         /// <summary>
--- a/src/Spatial/Euclidean/StraightLine2D.cs
+++ b/src/Spatial/Euclidean/StraightLine2D.cs
@@ -1,0 +1,315 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// Represents a line conforming to the equation y = mx + b where m is the gradient and b is the y-intercept
+    /// </summary>
+    public struct StraightLine2D : ILine
+    {
+        private readonly double yoffset;
+        private readonly double xoffset;
+
+        /// <summary>
+        /// The gradient of the line
+        /// </summary>
+        public double Gradient { get; }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b where point is a point on the line and gradient is the slope of the line
+        /// </summary>
+        /// <param name="point">A point on the line</param>
+        /// <param name="gradient">The gradient of the line</param>
+        public StraightLine2D(Point2D point, double gradient)
+        {
+            Gradient = gradient;
+            if (double.IsInfinity(Gradient))
+            {
+                yoffset = double.NaN; // special case vertical line has no y-intercept
+                xoffset = point.X;
+            }
+            else
+            {
+                yoffset = point.Y - (Gradient * point.X);
+                xoffset = double.NaN;
+            }
+        }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b which passes through points p1 and p2
+        /// </summary>
+        /// <param name="p1">A point on the line</param>
+        /// <param name="p2">A point on the line</param>
+        public StraightLine2D(Point2D p1, Point2D p2)
+        {
+            if (p1 == p2)
+            {
+                throw new ArgumentException("A line cannot be formed from two identical points");
+            }
+
+            Gradient = (p2.Y - p1.Y) / (p2.X - p1.X);
+            if (double.IsInfinity(Gradient))
+            {
+                yoffset = double.NaN; // special case vertical line has no y-intercept
+                xoffset = p1.X;
+            }
+            else
+            {
+                yoffset = p1.Y - (Gradient * p1.X);
+                xoffset = double.NaN;
+            }
+        }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b which passes through the points provided
+        /// </summary>
+        /// <param name="points">The points though which the line should pass</param>
+        /// <exception cref="ArgumentException">Thrown when less than 2 points are provided</exception>
+        /// <exception cref="ArgumentException">Thrown when points are not collinear</exception>
+        public StraightLine2D(IEnumerable<Point2D> points) : this(points.ToArray())
+        {
+        }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b which passes through the points provided
+        /// </summary>
+        /// <param name="points">The points though which the line should pass</param>
+        /// <exception cref="ArgumentException">Thrown when less than 2 points are provided</exception>
+        /// <exception cref="ArgumentException">Thrown when points are not collinear</exception>
+        public StraightLine2D(Point2D[] points)
+        {
+            if (points.Length < 2)
+            {
+                throw new ArgumentException("A line must be constructed from at least two points");
+            }
+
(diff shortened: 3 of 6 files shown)


src/Spatial/Euclidean/ILine.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// This interface represents a line in cartesian space
+    /// </summary>
+    public interface ILine : IEquatable<ILine>
```
JohanLarsson: Name it `ILine2D` until we have decided on namespaces.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// Represents a line conforming to the equation y = mx + b where m is the gradient and b is the y-intercept
+    /// </summary>
+    public struct StraightLine2D : ILine
```
JohanLarsson: I think we want ` : IEquatable<StraightLine2D>` alo here to avoid boxing.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b which passes through the points provided
+        /// </summary>
+        /// <param name="points">The points though which the line should pass</param>
+        /// <exception cref="ArgumentException">Thrown when less than 2 points are provided</exception>
+        /// <exception cref="ArgumentException">Thrown when points are not collinear</exception>
+        public StraightLine2D(IEnumerable<Point2D> points) : this(points.ToArray())
```
JohanLarsson: Not sure we should have this ctor, I know you added it for consistency but maybe we should go the other way and deprecate the other places?

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b which passes through the points provided
+        /// </summary>
+        /// <param name="points">The points though which the line should pass</param>
+        /// <exception cref="ArgumentException">Thrown when less than 2 points are provided</exception>
+        /// <exception cref="ArgumentException">Thrown when points are not collinear</exception>
+        public StraightLine2D(Point2D[] points)
```
JohanLarsson: Not sure we should have this ctor, I know you added it for consistency but maybe we should go the other way and deprecate the other places?

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+            return double.IsInfinity(Gradient);
+        }
+
+        /// <summary>
+        /// Tests if the other line is parallel to this one
+        /// </summary>
+        /// <returns>Returns true if lines are parallel</returns>
+        public bool IsParallel(ILine otherLine)
+        {
+            if (otherLine is StraightLine2D)
```
JohanLarsson: This does not look nice.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+        public bool IsVertical()
+        {
+            return double.IsInfinity(Gradient);
+        }
+
+        /// <summary>
+        /// Tests if the other line is parallel to this one
+        /// </summary>
+        /// <returns>Returns true if lines are parallel</returns>
+        public bool IsParallel(ILine otherLine)
```
JohanLarsson: Making it `public bool IsParalle<T>l(T otherLine) where T : ILine` avoids boxing.

src/Spatial/Euclidean/ILine.cs
AQuentin2033: ok

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+                    throw new ArgumentException("The points provided are not collinear");
+                }
+            }
+        }
+
+        /// <summary>
+        /// Determines if the line is horizontal
+        /// </summary>
+        /// <returns>True if the line is horizontal</returns>
+        public bool IsHorizontal()
```
JohanLarsson: Method or property? I don't remember how we have done in other places but feels like a property.

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: ok

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Tests if the other line is parallel to this one
+        /// </summary>
+        /// <returns>Returns true if lines are parallel</returns>
+        public bool IsParallel(StraightLine2D otherLine)
```
JohanLarsson: Should probably have a tolerance parameter, not sure what a good default is. An overload with `Angle tolerance` can probably make sense.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        /// <summary>
+        /// Finds the midpoint of two points along the current line
+        /// </summary>
+        /// <param name="pointA">a point on the line</param>
+        /// <param name="pointB">a point on the line</param>
+        /// <returns>A point midway between the two provided points</returns>
+        public Point2D MidPoint(Point2D pointA, Point2D pointB)
+        {
+            return new Point2D((pointA.X + pointB.X) / 2, (pointA.Y + pointB.Y) / 2);
```
JohanLarsson: How does this find the midpoint 'along the current line'? Also there is `Point2D.MidPoint()` IIRC.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        /// <summary>
+        /// Returns the list of points which lie on a horizontal line at coordinate y
+        /// </summary>
+        /// <param name="y">the y coordinate</param>
+        /// <returns>A list of points</returns>
+        public IEnumerable<Point2D> YIntercept(double y = 0)
+        {
+            return new List<Point2D>() { new Point2D((y - yoffset) / Gradient, y) };
```
JohanLarsson: I prefer `return new Point2D>[] { new Point2D((y - yoffset) / Gradient, y) };` here but not important.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+            return new List<Point2D>() { new Point2D(x, (Gradient * x) + yoffset) };
+        }
+
+        /// <summary>
+        /// Returns the list of points which lie on a horizontal line at coordinate y
+        /// </summary>
+        /// <param name="y">the y coordinate</param>
+        /// <returns>A list of points</returns>
+        public IEnumerable<Point2D> YIntercept(double y = 0)
```
JohanLarsson: Maybe this should be explicit interface implementation, dunno.

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: Wouldn't people want to construct a line from a list of points?

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: ok

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: I'm ok with changing that, a fixed array of items is perhaps better in this case than the enumerable one

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: midpoint formula - http://www.purplemath.com/modules/midpoint.htm
it so happens that the line is straight - if it was curved this would be different - and yes I know its not part of the ILine2D interface yet, it will be eventually.

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: Currently the internal implementation uses doubles to store the gradients but it is preferable that longer term that this stores the line internally as an equation using Mathnet.symbolics.  This would avoid the need for a tolerance and I think it would be better not to add to the API now only to remove it later

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: either or works.  In the current implementation the code is mostly trivial, even if we switch to using formulas internally this would still be fairly trivial.

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: because?

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: changing ctors to factory methods Create, CreateFromCollinear

src/Spatial/Euclidean/StraightLine2D.cs
JohanLarsson: But there is nothing saying the midpoint is on the line.

src/Spatial/Euclidean/StraightLine2D.cs
JohanLarsson: `IReadOnlyList<Point2D>` is nice, shows that there is no lazy stuff going on. And readonly.

src/Spatial/Euclidean/StraightLine2D.cs
JohanLarsson: Returning `IEnumerable<Point2D>` feels weird for a straight line. It has either one, zero or infinite. For some curved thing it makes more sense I think.

JohanLarsson: Maybe explicit implementation and a `public bool TryGetYIntercept(out Point2D)`

src/Spatial/Euclidean/ILine2D.cs
```diff
@@ -8,21 +8,21 @@ namespace MathNet.Spatial.Euclidean
...
-    public interface ILine : IEquatable<ILine>
+    public interface ILine2D : IEquatable<ILine2D>
     {
         /// <summary>
         /// Returns the points at which the line crosses on the x axis
         /// </summary>
         /// <param name="x">An optional x coordinate</param>
         /// <returns>A list of crossing points</returns>
-        IEnumerable<Point2D> XIntercept(double x = 0);
+        Point2D[] XIntercept(double x = 0);
```
JohanLarsson: Prefer `IReadOnlyList<Point2D>`

src/Spatial/Euclidean/ILine2D.cs
```diff
@@ -44,6 +44,6 @@ public interface ILine : IEquatable<ILine>
         /// </summary>
         /// <param name="line">the line to be compared against</param>
         /// <returns>True if the lines are parallel</returns>
-        bool IsParallel(ILine line);
+        bool IsParallel<T>(T line) where T : ILine2D;
```
JohanLarsson: I still think `bool IsParallel<T>(T line, Angle tolerance)` make sense. But we can wait and see what other people have to say.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -79,61 +60,65 @@ public StraightLine2D(IEnumerable<Point2D> points) : this(points.ToArray())
...
         /// Tests if the other line is parallel to this one
         /// </summary>
         /// <returns>Returns true if lines are parallel</returns>
-        public bool IsParallel(ILine otherLine)
+        public bool IsParallel<T>(T otherLine) where T : ILine2D
         {
             if (otherLine is StraightLine2D)
             {
-                IsParallel((StraightLine2D)otherLine);
+                IsParallel(otherLine);
```
JohanLarsson: Can remove the if now right?

AQuentin2033: if its another type of line say comparing a straightline with a parabola, which is a valid thing to do but should be false, so I think the if has to remain.

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: good point - will fix

src/Spatial/Euclidean/StraightLine2D.cs
JohanLarsson: Ok, I'm not sure `ILine2D` should have this method then but I don't know your plans for it yet.

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,312 @@
...
+            return CreateFromCollinear(points.ToArray());
+        }
+
+        /// <summary>
+        /// Creates a Line conforming to the equation y = mx + b which passes through the points provided
+        /// </summary>
+        /// <param name="points">The points though which the line should pass</param>
+        /// <exception cref="ArgumentException">Thrown when less than 2 points are provided</exception>
+        /// <exception cref="ArgumentException">Thrown when points are not collinear</exception>
+        public static StraightLine2D CreateFromCollinear(Point2D[] points)
```
JohanLarsson: this can be params Point2D[] points.

src/Spatial/Euclidean/StraightLine2D.cs
JohanLarsson: In other places we have named it `IsParallelTo` https://github.com/mathnet/mathnet-spatial/blob/master/src/Spatial/Euclidean/Line2D.cs#L182

src/Spatial/Euclidean/StraightLine2D.cs
```diff
@@ -0,0 +1,312 @@
...
+        {
+            return StraightLine2D.Create(p, -1 / Gradient);
+        }
+
+        /// <summary>
+        /// Find the intersection between two lines
+        /// </summary>
+        /// <param name="line">The line to intersect with</param>
+        /// <returns>The point of intersection or null if the lines are parallel</returns>
+        public Point2D? Intersection(StraightLine2D line)
```
JohanLarsson: I think we should deprecate the Intersection that returns null in other parts of the lib and make them `public bool TryIntersectWith(StraightLine2D line, out Point2D result)` what do you think about this?

src/Spatial/Euclidean/StraightLine2D.cs
milleniumbug: Who's the end user of the library? Since not everyone can afford symbolic computations, I assume (for example) that game developers aren't part of the user base.

src/Spatial/Euclidean/ILine2D.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// This interface represents a line in cartesian space
+    /// </summary>
+    public interface ILine2D : IEquatable<ILine2D>
+    {
+        /// <summary>
+        /// Returns the points at which the line crosses on the x axis
+        /// </summary>
+        /// <param name="x">An optional x coordinate</param>
+        /// <returns>A list of crossing points</returns>
+        Point2D[] XIntercept(double x = 0);
```
milleniumbug: Don't return an array. `IEnumerable<Point2D>` is preferable because it can be precomputed or cached or calculated lazily or eagerly.

AQuentin2033: this is what I had ... Johan asked to change it...

src/Spatial/Euclidean/ILine2D.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MathNet.Spatial.Euclidean
+{
+    /// <summary>
+    /// This interface represents a line in cartesian space
+    /// </summary>
+    public interface ILine2D : IEquatable<ILine2D>
```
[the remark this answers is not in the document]
AQuentin2033: Parabola and Hyperbola were interesting to add.  then if I can rope in mathnet symbolics a generic curve class

src/Spatial/Euclidean/StraightLine2D.cs
AQuentin2033: not a huge fan to be honest, breaks the flow of using it

PR #70, opened 2017-12-02
Fix StyleCop warnings

JohanLarsson:
Keept stylecop settings at default. This PR is mostly fixes of whitespace nags.

### DIFF
--- a/src/Spatial/Euclidean/PolyLine2D.cs
+++ b/src/Spatial/Euclidean/PolyLine2D.cs
@@ -48,7 +48,10 @@
         {
             double length = 0;
             for (int i = 0; i < this.points.Count - 1; ++i)
+            {
                 length += this[i].DistanceTo(this[i + 1]);
+            }
+
             return length;
         }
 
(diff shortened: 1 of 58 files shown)


src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -27,18 +26,18 @@ public class PolyLine2D : IEnumerable<Point2D>
         /// </summary>
         /// <param name="key"></param>
         /// <returns></returns>
-        public Point2D this[int key] => this._points[key];
+        public Point2D this[int key] => this.points[key];
 
         // Internal storage for the points
-        private List<Point2D> _points;
+        private readonly List<Point2D> points;
```
AQuentin2033: this is a behavior change - it wasn't readonly before and the collection was exposed publically via the indexer

src/Spatial/Euclidean/PolyLine3D.cs
```diff
@@ -27,18 +26,18 @@ public class PolyLine3D : IEnumerable<Point3D>
         /// </summary>
         public bool IsPlanar
         {
-            get { throw new NotImplementedException();}
+            get { throw new NotImplementedException(); }
         }
 
-        private List<Point3D> _points;
+        private readonly List<Point3D> points;
```
AQuentin2033: also behavior change making it readonly

src/Spatial/Euclidean/PolyLine2D.cs
JohanLarsson: The indexer is get only, how do you mean it is breaking?

PR #77, opened 2017-12-06
ImmutableList for internal use.

JohanLarsson:


src/Spatial/Internals/ImmutableList.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace MathNet.Spatial.Internals
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    internal sealed class ImmutableList<T> : IEnumerable<T>
```
JohanLarsson: Not sure about adding `IEnumerable<T>` here as it invites use of LINQ.

src/Spatial/Internals/ImmutableList.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        internal static readonly ImmutableList<T> Empty = new ImmutableList<T>();
+
+        private readonly T[] data;
+
+        private ImmutableList()
+        {
+            this.data = new T[0];
```
milleniumbug: the empty instance can be shared, not sure if it's a good idea in this case

src/Spatial/Internals/ImmutableList.cs
AQuentin2033: linq in intself is not evil - sometimes its even efficient, and with the new benchmarking we should catch poor implementations better

src/Spatial/Internals/ImmutableList.cs
JohanLarsson: Not sure what can happen with an empty array. We can remove that ctor and use the other ctor and pass in an empty list, reduces noise a bit.

src/Spatial/Internals/ImmutableList.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    internal sealed class ImmutableList<T> : IEnumerable<T>
+    {
+        internal static readonly ImmutableList<T> Empty = new ImmutableList<T>();
+
+        private readonly T[] data;
```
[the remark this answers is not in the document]
AQuentin2033: I think private we are protecting library developers against accidental changes

src/Spatial/Internals/ImmutableList.cs
AQuentin2033: there could be cases where it would be perfectly valid to return an empty list

src/Spatial/Internals/ImmutableList.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private ImmutableList()
+        {
+            this.data = new T[0];
+        }
+
+        internal ImmutableList(T[] data)
```
milleniumbug: so... we're relying that the user doesn't keep the argument around? worth documenting if that's the case

src/Spatial/Internals/ImmutableList.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public int Count => this.data.Length;
+
+        public T this[int index] => this.data[index];
+
+        public IEnumerator<T> GetEnumerator() => ((IList<T>)this.data).GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => this.data.GetEnumerator();
+
+        internal ImmutableList<T> Add(T value)
```
milleniumbug: I assume most of the lists will be small and nobody will be doing `Add` and `Remove` in the loop

src/Spatial/Internals/ImmutableList.cs
JohanLarsson: ok, keeping it for now then, can always be refactored later.

src/Spatial/Internals/ImmutableList.cs
JohanLarsson: We can add an AddRange method

src/Spatial/Internals/ImmutableList.cs
JohanLarsson: Yes, maybe better to make this private and have the internal ctor take an `IEnumerable<T>` and ToArray it.

PR #80, opened 2017-12-07
fix for polygon edges + test cases

AQuentin2033:
Fixed wrong edge implementation + added test case

### DIFF
--- a/src/Spatial/Euclidean/Polygon2D.cs
+++ b/src/Spatial/Euclidean/Polygon2D.cs
@@ -456,16 +456,14 @@
         /// </summary>
         private void PopulateEdgeList()
         {
-            var localedges = new List<Line2D>();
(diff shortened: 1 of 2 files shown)


src/Spatial/Euclidean/Polygon2D.cs
```diff
@@ -456,16 +456,14 @@ private static void RecursiveHullComputation(Point2D a, Point2D b, List<Point2D>
         /// </summary>
         private void PopulateEdgeList()
         {
-            var localedges = new List<Line2D>();
-            for (var i = 0; i < this.points.Count; i++)
+            var localedges = new List<Line2D>(this.points.Count);
```
JohanLarsson: should be `(this.points.Count - 1)` I think.

PR #87, opened 2017-12-07
Stylecop fixes Polyline3D

AQuentin2033:
General comments and cleanup

deleted not implemented method - IsPlanarWithinTol

### DIFF
--- a/src/Spatial/Euclidean/PolyLine3D.cs
+++ b/src/Spatial/Euclidean/PolyLine3D.cs
@@ -10,25 +10,33 @@
     /// </summary>
     public class PolyLine3D : IEnumerable<Point3D>
     {
+        /// <summary>
+        /// An internal list of points
+        /// </summary>
         private readonly List<Point3D> points;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PolyLine3D"/> class.
+        /// Creates a PolyLine3D from a pre-existing IEnumerable of Point3Ds
+        /// </summary>
+        /// <param name="points">A list of points.</param>
         public PolyLine3D(IEnumerable<Point3D> points)
         {
             this.points = new List<Point3D>(points);
         }
 
         /// <summary>
-        /// An integer representing the number of Point3D objects in the polyline
+        /// Gets an integer representing the number of Point3D objects in the polyline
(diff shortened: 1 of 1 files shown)


src/Spatial/Euclidean/PolyLine3D.cs
```diff
@@ -10,58 +10,53 @@
...
+        /// </summary>
         private readonly List<Point3D> points;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PolyLine3D"/> class.
+        /// Creates a PolyLine3D from a pre-existing IEnumerable of Point3Ds
+        /// </summary>
+        /// <param name="points">A list of points.</param>
         public PolyLine3D(IEnumerable<Point3D> points)
         {
```
JohanLarsson: We could add a null check here, throwexpression perhaps. Just a reminder, no need to fix in this PR.

src/Spatial/Euclidean/PolyLine3D.cs
```diff
@@ -10,58 +10,53 @@
...
+        /// Gets the length of the polyline, computed as the sum of the lengths of every segment
         /// </summary>
         public double Length => this.GetPolyLineLength();
 
         /// <summary>
-        /// Indicates whether or not the collection of points in the polyline are planar within
+        /// Gets a value indicating whether or not the collection of points in the polyline are planar within
         /// the floating point tolerance
         /// </summary>
         public bool IsPlanar
```
JohanLarsson: We should remove this or write an implementaion, I lean towards removing it.

src/Spatial/Euclidean/PolyLine3D.cs
AQuentin2033: this is just the doco pr, haven't yet started tuning the class

src/Spatial/Euclidean/PolyLine3D.cs
```diff
@@ -133,20 +128,31 @@ public Point3D ClosestPointTo(Point3D p)
             return closest;
         }
 
-        public bool IsPlanarWithinTol(double tolerance)
```
JohanLarsson: Was this unused & removed?

PR #106, opened 2017-12-09
Ray3D documentation + Obsolete Parser

AQuentin2033:
Ray3D documentation
Obsolete static Parse call to already obsoleted method

### DIFF
--- a/src/Spatial/Euclidean/Ray3D.cs
+++ b/src/Spatial/Euclidean/Ray3D.cs
@@ -98,6 +98,7 @@ namespace MathNet.Spatial.Euclidean
         /// </summary>
         /// <param name="s">a string representing the ray</param>
         /// <returns>a ray</returns>
+        [Obsolete("Should not have been made public. Mode Obsolete 2017-12-09")]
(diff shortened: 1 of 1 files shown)


src/Spatial/Euclidean/Ray3D.cs
```diff
@@ -98,6 +98,7 @@ public static Ray3D Parse(string point, string direction)
         /// </summary>
         /// <param name="s">a string representing the ray</param>
         /// <returns>a ray</returns>
+        [Obsolete("Should not have been made public. Mode Obsolete 2017-12-09")]
```
JohanLarsson: I'm not sure it must be obsoleted, parsing can be nice in unit tests.

AQuentin2033: it calls an obsolete method - isn't it enough to construct the fundamental types (points, vectors, etc) for the unittests without having another custom non standard string format?

JohanLarsson: ok, merging it

PR #109, opened 2017-12-10
Fix #63

### LINKED ISSUE #63: Point3D references in Point2D
Logically a 2D point shouldn't have knowledge of higher dimensions.  If a library user wishes to convert a 2D point into a 3D point it is more likely they would wish to project the point onto a plane in 3D space which is not handled by this function.

I therefor propose to obsolete the method ToPoint3D method and the Point3D operators.

AQuentin2033:
Obsolete the methods on Point2D which return Point3D - Projections should be used instead.


JohanLarsson:
Yes, this is probably the sensible thing to do.

Issue #99, opened 2017-12-09
Naming of Line2D/Line3D


AQuentin2033:
Suggest we rename Line2D/Line3D to LineSegment2D and LineSegment3D to better reflect their use

JohanLarsson:
Yes, we should probably pull the trigger on this. At least if the analyzer + fix for breaking changes turns out ok. It will not handle transitive dependencies but don't expect many libs to reference this given prerelease.

PR #112, opened 2017-12-10
New LineSegment2D

AQuentin2033:
Partial fix for issue #99.

Most methods are copied over from Line2D.  Changes are:
Only kept comparison operator overloads - other operators not carried over
new equals overload with tolerance
new method Translate
IsParallel tolerance only method kept
Intersection changed to TryIntersect with tolerance

Should any other methods be included?

### DIFF
--- a/src/Spatial/Euclidean/LIneSegment2D.cs
+++ b/src/Spatial/Euclidean/LIneSegment2D.cs
@@ -30,7 +30,7 @@
         {
             if (startPoint == endPoint)
             {
-                throw new ArgumentException("The Line2D starting and ending points cannot be identical");
+                throw new ArgumentException("The segment starting and ending points cannot be identical");
             }
 
             this.StartPoint = startPoint;
@@ -88,7 +88,7 @@
         /// </summary>
         /// <param name="vector">A vector to apply</param>
         /// <returns>A new translated linesegment</returns>
-        public LineSegment2D Translate(Vector2D vector)
+        public LineSegment2D TranslateBy(Vector2D vector)
         {
             var startVector = this.StartPoint.ToVector2D().Add(vector);
             var endVector = this.EndPoint.ToVector2D().Add(vector);
--- a/src/Spatial/Euclidean/LIneSegment2D.cs
+++ b/src/Spatial/Euclidean/LIneSegment2D.cs
@@ -1,0 +1,224 @@
+﻿namespace MathNet.Spatial.Euclidean
+{
+    using System;
+    using System.Diagnostics.Contracts;
+    using MathNet.Spatial.Units;
+
+    /// <summary>
+    /// This structure represents a line between two points in 2-space.  It allows for operations such as
+    /// computing the length, direction, comparisons, and shifting by a vector.
+    /// </summary>
+    public struct LineSegment2D : IEquatable<LineSegment2D>
+    {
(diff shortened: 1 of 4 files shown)


src/Spatial/Euclidean/LIneSegment2D.cs
```diff
@@ -0,0 +1,224 @@
...
+        /// Gets the distance from <see cref="StartPoint"/> to <see cref="EndPoint"/>
+        /// </summary>
+        [Pure]
+        public double Length => this.StartPoint.DistanceTo(this.EndPoint);
+
+        /// <summary>
+        /// Gets a normalized vector in the direction from <see cref="StartPoint"/> to <see cref="EndPoint"/>
+        /// </summary>
+        [Pure]
+        public Vector2D Direction => this.StartPoint.VectorTo(this.EndPoint).Normalize();
```
JohanLarsson: You really hate `UnitVector3D` :) I think this should be `UnitVector3D` as long as it is around.

src/Spatial/Euclidean/LIneSegment2D.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            return new LineSegment2D(Point2D.Parse(startPointString), Point2D.Parse(endPointString));
+        }
+
+        /// <summary>
+        /// Translates a line according to a provided vector
+        /// </summary>
+        /// <param name="vector">A vector to apply</param>
+        /// <returns>A new translated linesegment</returns>
+        public LineSegment2D Translate(Vector2D vector)
```
JohanLarsson: We can use TranslateBy to maximize sameness, think it is named with suffix By in other parts of the lib.

src/Spatial/Euclidean/LIneSegment2D.cs
```diff
@@ -0,0 +1,224 @@
...
+        /// <summary>
+        /// Compute the intersection between two lines if the angle between them is greater than a specified
+        /// angle tolerance.
+        /// </summary>
+        /// <param name="other">The other line to compute the intersection with</param>
+        /// <param name="intersection">When this method returns, contains the intersection point, if the conversion succeeded, or the default point if the conversion failed.</param>
+        /// <param name="tolerance">The tolerance used when checking if the lines are parallel</param>
+        /// <returns>True if an intersection exists; otherwise false</returns>
+        [Pure]
+        public bool TryIntersect(LineSegment2D other, out Point2D intersection, Angle tolerance)
```
JohanLarsson: We can probably add an Intersect method that throws on failure that calls this internally later. But that is another PR and we need to bikeshed it some due to existing Intersect methods returning `Nullable<Point2D>`

src/Spatial/Euclidean/LIneSegment2D.cs
AQuentin2033: we said that before...if only I could remember ... changing it :)

src/Spatial/Euclidean/LIneSegment2D.cs
AQuentin2033: there isn't a UnitVector2D ...

JohanLarsson: oh, derp, my bad, sry

PR #113, opened 2017-12-10
LineSegment3D

AQuentin2033:
Second part of fix towards #99 

Class mostly implements the same interface as Line3D with some exceptions:
Functions which dealt with pure lines were altered to deal with segments only
Added TranslateBy
replaced ClosestPointsBetween by TryShortestLineTo
added equals with tolerance
Only the IsParallelTo with tolerance was retained
Projections and intersections with planes are not included for now

### DIFF
--- a/src/Spatial/Euclidean/LIneSegment3D.cs
+++ b/src/Spatial/Euclidean/LIneSegment3D.cs
@@ -1,0 +1,317 @@
+﻿namespace MathNet.Spatial.Euclidean
+{
+    using System;
+    using System.Diagnostics.Contracts;
+    using MathNet.Spatial.Units;
+
+    /// <summary>
+    /// This structure represents a line between two points in 3D-space.  It allows for operations such as
+    /// computing the length, direction, comparisons, and shifting by a vector.
+    /// </summary>
+    public struct LineSegment3D : IEquatable<LineSegment3D>
+    {
+        /// <summary>
+        /// The starting point of the line segment
+        /// </summary>
+        public readonly Point3D StartPoint;
+
+        /// <summary>
+        /// The end point of the line segment
+        /// </summary>
+        public readonly Point3D EndPoint;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LineSegment3D"/> struct.
+        /// Throws an ArgumentException if the <paramref name="startPoint"/> is equal to the <paramref name="endPoint"/>.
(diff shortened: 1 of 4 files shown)


src/Spatial/Euclidean/LIneSegment3D.cs
```diff
@@ -0,0 +1,317 @@
...
+        }
+
+        /// <summary>
+        /// Translates a line according to a provided vector
+        /// </summary>
+        /// <param name="vector">A vector to apply</param>
+        /// <returns>A new translated linesegment</returns>
+        public LineSegment3D TranslateBy(Vector3D vector)
+        {
+            var startVector = this.StartPoint.ToVector3D() + vector;
```
JohanLarsson: Point3D + Vector3D = Point3D

src/Spatial/Euclidean/LIneSegment3D.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// <summary>
+        /// Computes the pair of points which represent the closest distance between this Line3D and another Line3D, with the first
+        /// point being the point on this Line3D, and the second point being the corresponding point on the other Line3D.  If the lines
+        /// intersect the points will be identical, if the lines are parallel the first point will be the start point of this line.
+        /// </summary>
+        /// <param name="other">line to compute the closest points with</param>
+        /// <param name="tolerance">A tolerance (epsilon) to adjust for floating point error</param>
+        /// <returns>A tuple of two points representing the endpoints of the shortest distance between the two lines</returns>
+        [Pure]
+        private Tuple<Point3D, Point3D> ClosestPointsBetweenLines(LineSegment3D other, Angle tolerance)
```
JohanLarsson: Not sure what is the best signature here, we can merge it as is as it is a port of Line3D.

src/Spatial/Euclidean/LIneSegment3D.cs
AQuentin2033: good point, will simplify

src/Spatial/Euclidean/LIneSegment3D.cs
AQuentin2033: if we add a replacement Line2D class in the future which is not bounded by points, then this method would go away in favor of using whatever the public interface of Line would be, hence I opted for just making it private and leaving it as it was

PR #116, opened 2017-12-11
Serialization tests

AQuentin2033:
Added some initial serialization tests for common approaches

BinaryFormatter
ProtoBuf-net
XmlSerializer
DataContract (Xml)
Json


src/SpatialUnitTests/Serialization/JsonTests.cs
```diff
@@ -0,0 +1,206 @@
+﻿namespace MathNet.Spatial.Serialization.Xml.UnitTests
```
JohanLarsson: Is this the correct namespace? Not very important.

PR #119, opened 2017-12-11
Try appveyor tests

AQuentin2033:
Appveyor Tests

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 image: Visual Studio 2017
 build_script:
-  - build.cmd
+  - build.cmd Build
 test:
   # only assemblies to test
   assemblies:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,6 @@ build_script:
 test: 
     Assemblies:
         Only:
-            - '**\MathNet.Spatial.UnitTests.dll
+            - '**\MathNet.Spatial.UnitTests.dll'
 deploy: off
 shallow_clone: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,10 @@
 image: Visual Studio 2017
 build_script:
   - build.cmd
-test: 
-    Assemblies:
-        Only:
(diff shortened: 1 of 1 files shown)


JohanLarsson:
@cdrnet Do we want fake to build on Appveyor?

cdrnet:
Yes, if feasible. It seems to build fine, including tests. Is it causing other problems?

cdrnet:
On a related note, Math.NET Numerics, which uses the same build infrastructure (as all Math.NET projects) is currently migrating its usage of msbuild towards dotnet sdk (in order to automate its .Net Standard support), but is still orchestrated using FAKE. But this is still work in progress.

cdrnet:
Btw, with the changes of this PR we're now running the tests twice on AppVeyor. Was this intended?

appveyor.yml
```diff
@@ -1,6 +1,9 @@
 image: Visual Studio 2017
 build_script:
-  - build.cmd Test
-test: off
+  - build.cmd
```
cdrnet: If we just want to build, without running tests, we can use `build.cmd Build`. Without argument it will run the default target, which is currently `Test` (but we should not rely on the default target here).

JohanLarsson:
@cdrnet Not a problem but it does not copy results from the test run. The AppVeyor UI for tests is pretty nice and useful ime.

cdrnet:
Yes, for some reason FAKE does not seem to pass the argument to the nunit runner to output appveyor messages. Either we use the task wrong, or this needs a fix in FAKE.

The output works for the last build of the PR, where AppVeyor controls the test run (with its own disadvantages): https://ci.appveyor.com/project/cdrnet/mathnet-spatial/build/1.0.253/tests

appveyor.yml
AQuentin2033: noted, will change the pr to only do the tests once (as was originally envisaged)

cdrnet:
Looks good to me know. Thanks!

PR #125, opened 2017-12-13
Tests for main build process

AQuentin2033:
Multitargeting for Mathnet.Spatial

Builds .netstandard 2.0, .netstandard 1.3 and .net 4.0 assemblies
using the dotnet build

Note: Currently references MathNet.Numerics 4.0.0-alpha02 (prerelease)


cdrnet:
Please don't use MathNet.Numerics.Core, this is an inofficial package. Version 4 of the official packages does support .Net Standard 1.3 and 2.0, although still prereleases only.

cdrnet:
(do we really need this right now, or can we wait for a couple months?)

AQuentin2033:
ok noted - just working my way through the build code.  I suppose it doesn't need to be immediately now - after all everything can always be done later ... but I do need it as .net standard at some point (for now I have just "borrowed" the code rather than using the nuget package but eventually I would prefer to use the package

cdrnet:
Of course, if you're already using it in a .Net Standard context we should not delay.

AQuentin2033:
of course that assumes I can actually make the thing work on appveyor and travis - works locally on windows both with VS and commandline without error so its not so obvious what the problem is.

cdrnet:
I guess on Linux the priority to run the tests on netcoreapp2.0 (against the netstandard build) is a bit higher than on mono (against net40 build). I've seen such issues with mono on such travis setups before, and we've had similar issues in Numerics (currently hidden by another issue).

AQuentin2033:
It now works for netcoreapp1.1 (.netstandard1.3 for library) and netcoreapp2.0 (netstandard2.0 library)
It also works on windows for net45 unit tests against a net40 library.

It broke the appveyor test ui integration (seems to have an issue with.netstandard)
It doesn't work yet for net45 on Travis ( I don't really understand how to configure this on Travis - so any help would be great)

JohanLarsson:
@cdrnet I'm thinking I'll merge this and we continue working on things, do you have any objections to that?

src/SpatialUnitTests/Euclidean/Ray3DTests.cs
```diff
@@ -56,20 +55,5 @@ public void XmlTests(string ps, string vs, bool asElements, string xml)
...
-            using (var ms = new MemoryStream())
-            {
-                var formatter = new BinaryFormatter();
-                formatter.Serialize(ms, v);
-                ms.Flush();
-                ms.Position = 0;
-                var roundTrip = (Ray3D)formatter.Deserialize(ms);
-                AssertGeometry.AreEqual(v, roundTrip);
-            }
-        }
```
cdrnet: Can't we keep the binary round trips in .Net Framework builds?

AQuentin2033: I'm not sure if you are thinking that I just removed them or you are objecting to me moving them.  They are still there, just they all live in BinaryFormatterTests.cs and are wrapped in an #if NETCOREAPP1_1 == false.  So they still run for both netcoreapp2.0 and net45.

If you are asking about me moving them then a few thoughts on this:
 - serialization as a set be it, binary, xml, json, protobuf, etc is a transversal concept.  If we support a particular serialization then it should work on all types in the library.  It is easier to track and ensure this by keeping the serialization pieces for a given target together.
- a given serialization is not particularly tied to the functionality of the type, it would be odd to have legacy binary serialization with the type but other serialization tests apart.  If they are all with the types then every unit test class would have non core dependencies
- having framework version specific support is easier to manage if there is a minimum of #if scattered through the code.  By grouping by serialization testing together, those checks can be skipped on mass for platforms where they are not supported

cdrnet:
No objections from my side!

src/SpatialUnitTests/Euclidean/Ray3DTests.cs
cdrnet: No, I was indeed missing the place - thanks for clarifying!

PR #127, opened 2017-12-14
Fix warnings.

JohanLarsson:


src/Spatial.Roslyn.Tests/ApiCompare/Reflector.cs
```diff
@@ -11,11 +10,7 @@ public class Reflector
         private Assembly legacy;
         private Assembly current;
 
-        private string oldAssemblyPath;
-        private string newAssemblyPath;
```
AQuentin2033: These are unused but the idea was to keep the paths, so that the assembly lookup could search their locations too - currently it only searches the current path - which is not really ideal

JohanLarsson: We add them back later when they are used?

AQuentin2033: sure

PR #128, opened 2017-12-15
Angle Formatter to fix #110

### LINKED ISSUE #110: IFormatProvider for Angle
Proposal to create a FormatProvider for Angle which can output angles in their various string formats
This would replace the use of Degrees and Radians classes and eliminate the final use of IAngleUnit

AQuentin2033:
Angle Formatter to fix #110

Similar to https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings for numbers with this change Angle supports its own format strings

D:  Formats the angle as degrees e.g. 59°
G:  Formats the angle in Gon or grad e.g. 30ᵍ
M:  Formats the angle in Mil format e.g. 1600 mil
N:  Formats the angle in Minutearc format e.g. 67.343411′
R:  Formats the angle as radians e.g. 34.554 rad
S:  Formats the angle in full Sexagesimal format e.g. 40° 11′ 15″

This is usable with 
`angle.ToString("D3").    // returns 30.222° ` 

A trailing number allows you to set precision.  So for example D3 shows degrees to 3 decimal places.
You can also qualify it with other formats supported by numbers such as:
`angle.ToString("D0.##").    // returns 30.22° ` 

Additionally the format provider can be used directly as in:
```
Angle angle = Angle.FromSexagesimal(30, 20, 10);
string.Format(new AngleFormatProvider(), "{0:S0}", angle); // returns 30° 20' 10"
```


src/Spatial/Units/Angle.cs
```diff
@@ -311,36 +311,39 @@ public static Angle ReadFrom(XmlReader reader)
...
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Returns a string representation of an Angle according to the default representation provided by <see cref="AngleFormatProvider"/>
+        /// </summary>
+        /// <returns>The string representation of this instance.</returns>
         public override string ToString()
         {
-            return this.ToString(null, NumberFormatInfo.CurrentInfo);
+            return this.ToString("R");
```
JohanLarsson: This will output 17 digits I think, if we change things we should be as close as what `double.ToString()` does as possible.

src/Spatial/Units/Angle.cs
```diff
@@ -311,36 +311,39 @@ public static Angle ReadFrom(XmlReader reader)
...
         }
 
         /// <summary>
-        /// Returns a string representation of the Angle using the provided format
+        /// Returns a string representation of the Angle using requested format.  See <see cref="AngleFormatProvider"/> for details of formatting options
         /// </summary>
-        /// <param name="format">a string indicating the desired format of the double.</param>
+        /// <param name="format">A string indicating the desired format</param>
```
JohanLarsson: I think the old version is clearer, maybe it can be made even clearer.

JohanLarsson:
I don't love this API as it is like how it is done in the framework.

src/Spatial/Units/Angle.cs
AQuentin2033: it returns 15 digits ... which is identical to double.ToString()

src/Spatial/Units/Angle.cs
AQuentin2033: the old version returned the format of the double - the new one returns the format of the angle - they are not the same thing in this implementation

AQuentin2033:
Currently when a user calls string.Format("{0}", angle)  It returns the ToString() value which is always <angle>nbsp;rad. 
With this extension the user can also call string.Format(new AngleFormatProvider(), "{0:D}", angle) and it will return the angle in degrees.

I'll agree that MS could have done this better in the framework and let types self-register but isn't it better for people to have the option?   Particularly if I add a couple more options to be able to format as Degrees, minutes, seconds and Gons
If they don't do anything the behavior is as it was prior to this extension, so no one loses anything

JohanLarsson:
What does it look like if I want to print the equivalent of `string.Format("{0:0.##}°", a.Degrees)`

AQuentin2033:
Added additional formats for Gons (Grads), Mil and minarc. 
Also obsoleted legacy Radians, Degrees and IUnit.
Added a static property with lazy contraction for the formatter - I think this offers users a better visual experience so users can now write

`string.Format(Angle.Formatter, "{0:D2}", a)`

rather than creating a new formatter each time,

I believe this change is now code complete.

AQuentin2033:
@JohanLarsson @cdrnet  I am not really ecstatic about this approach of leaving PRs languish until they are no longer easily mergable (due to other changes) if you don't like something about it.  I would much prefer if you would simply close it and say you are not interested in having that in the library.  I personally think its a nice enhancement but hey, its your library, not mine.

JohanLarsson:
It is our library, you are a significant contributor. You are right this could have been handled better but I was not ecstatic about the API and have been super busy with work, as always.

PR #137, opened 2018-01-11
Update build for MathNet.Numerics 4.0.0.Beta03

AQuentin2033:
Updated references to use latest mathnet.numerics nuget release
Sync build libraries with latest Numerics contents

@cdrnet  Only thing I didn't do is replace the csproj references for nuget with paket - going to leave that for someone else with more knowledge on paket to handle.


AQuentin2033:
So previously I had the build on travis not build the solution but rather build the test project only (which of course also builds its dependencies).  I have changed it to match numerics so that the build is performed once and then the test cases run against the already built dlls  However this breaks travis again looking for its SDKs.  Not sure yet what to do about this.

AQuentin2033:
I gave up and went back to the way it worked before.  Travis only builds .net core 1.1 and 2.0 as part of the test suite run and doesn't attempt to prebuild anything. 

This is ready to merge.

PR #140, opened 2018-01-13
Equality alignment for remaining entities

AQuentin2033:
All Euclidian types (+Angle) now consistently implement:
Equals
override Equals(object)
GetHashCode - using internal HashCode struct
Equals(T, double tolerance)
Operator overrides for == and !=

Additionally PolyLine2D and PolyLine3D are aligned to Polygon2D for vertex interface

### DIFF
--- a/src/Spatial/Euclidean/CoordinateSystem.cs
+++ b/src/Spatial/Euclidean/CoordinateSystem.cs
@@ -1,6 +1,7 @@
 namespace MathNet.Spatial.Euclidean
 {
     using System;
+    using System.Diagnostics.Contracts;
     using System.Linq;
     using System.Text.RegularExpressions;
     using System.Xml;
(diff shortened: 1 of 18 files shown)


src/Spatial/Euclidean/CoordinateSystem.cs
```diff
@@ -176,7 +177,7 @@ public CoordinateSystem BaseChangeMatrix
         /// <returns>True if the coordinate system are the same; otherwise false.</returns>
         public static bool operator ==(CoordinateSystem left, CoordinateSystem right)
         {
-            return CoordinateSystem.Equals(left, right);
+            return left.Equals(right);
```
JohanLarsson: Potential NRE here I think.

src/Spatial/Euclidean/CoordinateSystem.cs
```diff
@@ -742,61 +743,61 @@ public CoordinateSystem Invert()
...
-                return true;
-            }
-
             if (other.Values.Length != this.Values.Length)
             {
                 return false;
             }
 
-            return !this.Values.Where((t, i) => Math.Abs(other.Values[i] - t) > 1E-15).Any();
+            return !this.Values.Where((t, i) => Math.Abs(other.Values[i] - t) > tolerance).Any();
```
JohanLarsson: Any has an overload that takes a func

src/Spatial/Euclidean/CoordinateSystem.cs
AQuentin2033: consistency is consistency...

PR #146, opened 2018-01-16
Convenience members for Angle

AQuentin2033:
Adds the following:
Sin/Cos/Tan/Normalize methods to Angle
static angles for common angles 360, 180, 90

Explicit conversion from double to angle - allows (Angle)0.25 in the place of Angle.FromRadians(0.25).  This follows the logic that a double may be an angle but only if the programmer is sure it is.

### DIFF
--- a/src/Spatial/Units/Angle.cs
+++ b/src/Spatial/Units/Angle.cs
@@ -29,6 +29,11 @@ namespace MathNet.Spatial.Units
         public static readonly Angle Degrees360 = Angle.FromRadians(2 * Math.PI);
 
         /// <summary>
+        /// An angle representing 0 degrees or 0 radians
+        /// </summary>
+        public static readonly Angle Degrees0 = Angle.FromRadians(0);
+
+        /// <summary>
         /// The value in radians
         /// </summary>
         public readonly double Radians;
@@ -363,14 +368,14 @@ namespace MathNet.Spatial.Units
         }
 
         /// <summary>
-        /// Normalizes an angle to be constrained within start and end.
+        /// Clamps an angle to be constrained within start and end.
         /// </summary>
         /// <param name="start">The start angle</param>
         /// <param name="end">The end angle</param>
-        /// <returns>The normalized angle</returns>
-        public Angle Normalize(Angle start, Angle end)
+        /// <returns>The clamped angle</returns>
+        public Angle Clamp(Angle start, Angle end)
         {
-            return this.Normalize(start.Radians, end.Radians);
+            return this.Clamp(start.Radians, end.Radians);
         }
 
         /// <inheritdoc />
@@ -533,12 +538,12 @@ namespace MathNet.Spatial.Units
         }
 
         /// <summary>
-        /// Normalizes an angle to be constrained within start and end.
+        /// Clamps an angle to be constrained within start and end.
         /// </summary>
         /// <param name="start">The start angle in radians</param>
         /// <param name="end">The end angle in radians</param>
-        /// <returns>The normalized angle</returns>
-        internal Angle Normalize(double start, double end)
+        /// <returns>The clamped angle</returns>
+        internal Angle Clamp(double start, double end)
         {
             double width = end - start;
             double offset = this.Radians - start;
--- a/src/Spatial/Units/Angle.cs
+++ b/src/Spatial/Units/Angle.cs
@@ -14,6 +14,21 @@ namespace MathNet.Spatial.Units
     public struct Angle : IComparable<Angle>, IEquatable<Angle>, IFormattable, IXmlSerializable
     {
         /// <summary>
+        /// An angle representing 180 degrees or PI radians
+        /// </summary>
+        public static readonly Angle Degrees180 = Angle.FromRadians(Math.PI);
+
+        /// <summary>
+        /// An angle representing 90 degrees or PI / 2 radians
+        /// </summary>
+        public static readonly Angle Degrees90 = Angle.FromRadians(Math.PI / 2);
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Spatial/Units/Angle.cs
```diff
@@ -311,6 +335,44 @@ public static Angle ReadFrom(XmlReader reader)
...
+            return Math.Tan(this.Radians);
+        }
+
+        /// <summary>
+        /// Normalizes an angle to be constrained within start and end.
+        /// </summary>
+        /// <param name="start">The start angle</param>
+        /// <param name="end">The end angle</param>
+        /// <returns>The normalized angle</returns>
+        public Angle Normalize(Angle start, Angle end)
```
JohanLarsson: This is often called clamp

src/Spatial/Units/Angle.cs
```diff
@@ -68,6 +83,15 @@ private Angle(double radians)
         /// </summary>
         public double Degrees => this.Radians * RadToDeg;
 
+        /// <summary>
+        /// Explicit conversion from a numeric angle in radians to an Angle object
+        /// </summary>
+        /// <param name="radians">The angle in radians</param>
+        public static explicit operator Angle(double radians)
```
JohanLarsson: Not sure about this.

src/Spatial/Units/Angle.cs
```diff
@@ -13,6 +13,21 @@ namespace MathNet.Spatial.Units
     [Serializable]
     public struct Angle : IComparable<Angle>, IEquatable<Angle>, IFormattable, IXmlSerializable
     {
+        /// <summary>
```
JohanLarsson: No Zero?

src/Spatial/Units/Angle.cs
```diff
@@ -311,6 +335,44 @@ public static Angle ReadFrom(XmlReader reader)
             return reader.ReadElementAs<Angle>();
         }
 
+        /// <summary>
+        /// Gets the sine of the angle
+        /// </summary>
+        /// <returns>The sine of the angle</returns>
+        public double Sin()
```
JohanLarsson: Not sure about this `a.Cos()` reads a bit weird to me.

AQuentin2033: if I have created an angle, the idea that I have to turn it back into a double with myangle.Radians in order to get the sin or cos of the angle really doesn't encourage me to use Angle at all.  I start asking why does it exist?  Also from a simplicity point of view I find myangle.Sin() much more streamlined than Math.Sin(myangle.Radians);

src/Spatial/Units/Angle.cs
AQuentin2033: I started with the angles used in common geometry calculations.  I don't know that I have ever come across using a zero angle for anything, but why not

src/Spatial/Units/Angle.cs
AQuentin2033: if you have performed a calulcation involving angles it is odd to then need to go Angle.FromRadians(x) to turn it into an angle.  The double is an angle, logically I don't want to "construct" an angle, I have one, I just want the compiler to recognize it as what I say it is and offer me additional functionality that is available to the thing called "Angle"

src/Spatial/Units/Angle.cs
AQuentin2033: shrug I have heard both terms with regards to angles.  If you prefer Clamp I don't mind either way

src/Spatial/Units/Angle.cs
JohanLarsson: Angle is a very nice abstraction imo, I like types. When used as a parameter we make it explicit that it is an angle avoiding one class of bugs, passing the wrong thing. Also if we used doubles our parameter names would be `angleInRadians` and it would still be easy to make a mistake an pass in degrees.

`Math.Sin(myangle.Radians)` is nice and explicit imo. I don't know if extension everything in C#8 will allow us to add extension methods to static classes, if so adding overloads to `Math` taking `Angle`could be nice.

src/Spatial/Units/Angle.cs
milleniumbug: I'm not a fan either. There are multiple interpretations of what an explicit conversion from `double` to an `Angle` could mean (radians? degrees?), and promoting a single one is a great way to introduce unclear code.

src/Spatial/Units/Angle.cs
milleniumbug: Arguably `Degrees0` counts as zero

src/Spatial/Units/Angle.cs
AQuentin2033: I must admit I am confused that just above you agree that because there is an ambiguity between using degrees and using radians you prefer not to have explicit conversion but here you are comfortable with the fact that Math.Sin requires you to be aware that you need to pass in radians and not degrees and find this sufficient.

JohanLarsson: `Sin` as a member on angle has weird semantics to me. We can wait and see if other people have different opinion. Also I checked with the roslyn guys, looks like we will be able to add static overloads to `Math` that accepts our `Angle`

src/Spatial/Units/Angle.cs
AQuentin2033: The trig functions in the .net framework all use Radians.  In fact anywhere in the framework that you need to do something with an angle requires a single or double representing the angle in radians.  I would say that the boat promoting a single meaning has long since sailed in the .net world...

AQuentin2033:
Seems the consensus is no angle changes, its just perfect as it is

PR #147, opened 2018-01-16
Obsolete CoordinateSystem.Parse

AQuentin2033:
Moved the support parsing into unittesting where it belongs.  As of now no other unit tests make use of CoordinateSystem, so the functionality is local.  If that should change we can move this into helper class

Obsolete the public method and private member calls to already obsoleted Parser.


AQuentin2033:
Johan will handle this differently

PR #153, opened 2018-02-08
Update Numerics Reference to 4.0.0

### LINKED ISSUE #129: Travis CI doesn't currently build for mono
Currently Travis CI doesn't check the build against mono.

AQuentin2033:
Aligned build files to latest 'common' files from Mathnet.Numerics
Fixed compile warnings
Updated SDK and framework versions on Travis
Switched back to using Paket
Fix #129 - Linux build for mono is now working again
"Build Benchmarks" now correctly copies the results into the source directory


AQuentin2033:
@cdrnet  I have also synced the changes you have made to Numerics post Beta06 (until [this](https://github.com/mathnet/mathnet-numerics/commit/94477de4a07eb7724ba4bff9931d937447458720) commit ) .  If you have the time could you give me a brief description of the steps involved in preparing the signed version and in publishing so I can check those are working too?  Thanks

cdrnet:
Thanks a lot, this will make my life significantly easier in the near future!

The whole release process is fully automated, the only steps I usually need to do are:
* Update release notes file
* Run `./build.sh all sign`
* Create a release commit (expected local changes: the release notes, and the automatically updated csproj, AssemblyInfo)
* Run `./build.sh publish` if everything is ok

However, currently the publishing part in the last command is assuming that only I do releases, and thus does not even try to be generic. If you would like to do releases of Spatial as well (which would be great), I need to rethink this a bit.

Regarding signing, I assume you refer to the optional "sign" argument, not strong naming. I have a personal certificate which I technically cannot share (key is on a smart card), nor want to. But signing is not required, that's why it is a separate argument to opt-in. If you happen to have a code signing certificate as well, we'd need to make the build script a bit more flexible to allow choosing the right certificate.

Regarding testing, maybe it is easiest for now if I simply smoke-test it myself.

AQuentin2033:
Ok so based upon what you said - as far as I can tell "Build all" works and updates the csproj as expected, so feel free to merge this PR if you are happy to look after the sign/publish parts.

It currently builds as Spatial 0.5-Beta02, so you could reasonably run through the whole process including publishing to make sure it works.  It would at the very least be interesting to see how many people would download a preview version of spatial with .netstandard support.

AQuentin2033:
I updated this to the release version of Numerics 4.0.0.

JohanLarsson:
Did you switch to paket magic mode?
If yes paket.bootstrapper.exe should be removed.
If no paket.exe should not be in the repo as it is downloaded by the bootstrapper.

AQuentin2033:
Yes, Numerics switched to paket magic mode, so I did this also here and paket.bootstrapper.exe was removed in this PR.

PR #155, opened 2018-04-12
Update references + common build changes

Jones-Adam:
Updated references to Mathnet.Numerics4.4
Updated NewtonSoft.Json reference to 11.0
Merged in common build changes from Mathnet.Numerics4.4
temporary fix for benchmark paket references error
Expanded unit tests to include .net471


Jones-Adam:
@cdrnet @JohanLarsson - Sorry to say this but this project is feeling a bit abandoned lately...  I tried to inject some momentum but I think we have reached a kind of roadblock that only you two as owners can get past.  My own needs have long since diverged from what this library provides and while I would love to extend the library, I am not convinced that this makes sense in the current climate.  I realize that both of you probably have other priorities and that's fine, I just cannot see this library develop without active leadership. (of course, there is always the possibility that you might consider the library good enough as is, in which case that's ok, but then it would be nice if that was made clear so I and other contributors know what to expect :)

I really am sorry to be a nag... but after several thousands of lines of code, I find I care just a little ::P

PR #165, opened 2019-01-07
Add orthogonal to Vector2D

EMostafaAli:
Similar to what we have in Vector3D, I added Orthogonal for Vector2D

### DIFF
--- a/src/Spatial.Tests/Euclidean/Vector2DTests.cs
+++ b/src/Spatial.Tests/Euclidean/Vector2DTests.cs
@@ -196,6 +196,16 @@ namespace MathNet.Spatial.UnitTests.Euclidean
             Assert.AreEqual(2, actual[1]);
         }
 
+        [TestCase("1, 0")]
+        [TestCase("1, 1")]
+        [TestCase("1, -1")]
+        public void Orthogonal(string vs)
+        {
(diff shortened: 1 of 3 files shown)


src/Spatial.Tests/Euclidean/Vector2DTests.cs
```diff
@@ -196,6 +196,16 @@ public void ToDenseVector()
...
         }
 
+        [TestCase("1, 0")]
+        [TestCase("1, 1")]
+        [TestCase("1, -1")]
+        public void Orthogonal(string vs)
+        {
+            var v = Vector2D.Parse(vs);
+            var orthogonal = v.Orthogonal;
+            Assert.IsTrue(orthogonal.DotProduct(v) < 1e-6);
```
JohanLarsson: I prefer passing in a string with the expected vector and do `Assert.AreEqual(Vector2D.Parse(expected), orthogonal)` here.

EMostafaAli:
@JohanLarsson, I think it is nice to have it. Here is AutoCAD API showing the same implementation.
https://help.autodesk.com/view/OARX/2019/ENU/?guid=OREFNET-Autodesk_AutoCAD_Geometry_Vector3d_GetPerpendicularVector

cdrnet:
Thanks!

PR #167, opened 2019-04-19
Fix LineSegment2D.TryIntersect to return true only when segments inte…

imlex:
…rsect not lines.

### DIFF
--- a/src/Spatial.Tests/Euclidean/LineSegment2DTests.cs
+++ b/src/Spatial.Tests/Euclidean/LineSegment2DTests.cs
@@ -128,7 +128,7 @@
 
         [TestCase("0,0", "2,2", "1,0", "1,2", "1,1")]
         [TestCase("0,0", "2,2", "0,1", "2,1", "1,1")]
-        [TestCase("0,0", "2,2", "-1,-5", "-1,0", "-1,-1")]
+        [TestCase("-1,-1", "2,2", "-1,-5", "-1,0", "-1,-1")]
         public void IntersectWithTest(string s1, string e1, string s2, string e2, string expected)
         {
             var line1 = LineSegment2D.Parse(s1, e1);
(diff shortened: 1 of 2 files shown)


src/Spatial.Tests/Euclidean/LineSegment2DTests.cs
```diff
@@ -139,6 +139,19 @@ public void IntersectWithTest(string s1, string e1, string s2, string e2, string
             Assert.AreEqual(e, result);
         }
 
+        [TestCase("0,0", "-2,-2", "1,0", "1,2", "1,1")]
```
JohanLarsson: Why not add the test cases to the existing test?

src/Spatial.Tests/Euclidean/LineSegment2DTests.cs
```diff
@@ -128,7 +128,7 @@ public void ClosestPointToWithinSegment(string start, string end, string point,
 
         [TestCase("0,0", "2,2", "1,0", "1,2", "1,1")]
         [TestCase("0,0", "2,2", "0,1", "2,1", "1,1")]
-        [TestCase("0,0", "2,2", "-1,-5", "-1,0", "-1,-1")]
+        [TestCase("-1,-1", "2,2", "-1,-5", "-1,0", "-1,-1")]
```
JohanLarsson: We should add a test for parallel segments but that does not have to be in this PR.

src/Spatial/Euclidean/LineSegment2D.cs
```diff
@@ -157,9 +157,10 @@ public bool TryIntersect(LineSegment2D other, out Point2D intersection, Angle to
             var s = other.StartPoint.VectorTo(other.EndPoint);
 
             var t = (q - p).CrossProduct(s) / r.CrossProduct(s);
+            var u = (p - q).CrossProduct(r) / s.CrossProduct(r);
```
JohanLarsson: Why are both cross products needed?

src/Spatial.Tests/Euclidean/LineSegment2DTests.cs
imlex: Why not? One is for positive result, another is for negative.

PR #179, opened 2020-07-22
Update XmlExt.cs

rsinnig:
line 190: writer.WriteValue(value.ToString("G15")); ToString() without culture format provider parameter uses local culture. This causing issues when deserializing on machines with comma decimal separator.

### DIFF
--- a/src/Spatial/Obsolete/XmlExt.cs
+++ b/src/Spatial/Obsolete/XmlExt.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
--- a/src/Spatial/Obsolete/XmlExt.cs
+++ b/src/Spatial/Obsolete/XmlExt.cs
@@ -187,7 +187,7 @@ namespace MathNet.Spatial
         public static XmlWriter WriteAttribute(this XmlWriter writer, string name, double value)
         {
             writer.WriteStartAttribute(name);
-            writer.WriteValue(value.ToString("G15"));
+            writer.WriteValue(value.ToString("G15", CultureInfo.InvariantCulture));
             writer.WriteEndAttribute();
             return writer;
         }


JohanLarsson:
Ugh, this was really bad and in so many ways. This should not have been public and lack of `InvariantCulture` is embarrassing.

Thanks for this PR.
This is a breaking change.

src/Spatial/Obsolete/XmlExt.cs
```diff
@@ -187,7 +187,7 @@ public static XmlWriter WriteAttribute<T>(this XmlWriter writer, string name, T
         public static XmlWriter WriteAttribute(this XmlWriter writer, string name, double value)
         {
             writer.WriteStartAttribute(name);
-            writer.WriteValue(value.ToString("G15"));
+            writer.WriteValue(value.ToString("G15", CultureInfo.InvariantCulture));
```
JohanLarsson, line 190: CI fails, are you missing a using?

JohanLarsson:
`XmlExt` is `[Obsolete]` https://github.com/mathnet/mathnet-spatial/blob/379013b8280aca8a45147e3612fa6bd2e59f5519/src/Spatial/Obsolete/XmlExt.cs#L14

Not sure we should do a breaking change on an obsolete type.

rsinnig:
Sorry forgot to insert the using System.Globalization;

Actually I don't understand why it's marked obsolete but still in use.

But I had to modify this line on my local copy of code to get deserialization working.

Writing double values with comma decimal separator into XML serialisation is definitely not correct, since deserialization assumes culture neutral formatting by default.

JohanLarsson:
>Actually I don't understand why it's marked obsolete but still in use.

Are there still usages of this in the library?

>Writing double values with comma decimal separator into XML serialisation is definitely not correct, since deserialization assumes culture neutral formatting by default.

Agreed, this was a really unfortunate mistake.

rsinnig:
IXmlSerializable.WriteXml(XmlWriter writer) explicit implementation of Point2D, Point3D, UnitVector3D, Vector2D, Vector3D and Angle seems to use this extension method.

rsinnig:
Maybe I should cancel the pull request. The actual issue seems to be, that mentioned classes Point3D, … are using the obsolete methods instead of methods inside MathNet.Spatial.Internals.XmlWriterExt class.
I will try to fix it by using MathNet.Spatial.Internals.XmlWriterExt.WriteElement(this XmlWriter writer, string name, double value, string format)
But I can't do that work in this week.

rsinnig:
Cancelling of request to be done from my side?
It's my first PR on GitHub. I'm not experienced, with the usual process.

JohanLarsson:
Yes I don't think fixing this in the obsolete method is ideal so closing this pull request is probably good. Thank you for taking the time to contribute an d what you found is very troubling.

JohanLarsson:
> Cancelling of request to be done from my side?

I closed it, but could have been done from either side.

PR #181, opened 2020-07-29
Replace call to obsolete XmlExt by XmlWriteExt

rsinnig:
As discussed in cancelled pull request for fix in obsolete XmExt:
Fix of the XML serialization issue (writing without InvariantCulture) by replacing all calls to obsolete extension method XmlExt.WriteAttribute by XmlWriteExt.WriteElement.


rsinnig:
Failed test caused by different XML-formatting of obsolete XmlExt.WriteAttribute and new internal XmlWriteExt.WriteElementString
While XmlExt.WriteAttribute is writing start tag, value, end tag (no new line and tabs inserted) the XmlWriteExt.WriteElementString is creating nice formatting with newline and tabs as long as XmlWriter is not explicitly configured to omit these.

From my perspective the XmlSerializer test needs to be modified.

rsinnig:
The test procedure public static T XmlSerializerRoundTrip<T>(T item, string expected) uses the extension methods from obsolete XmlExt.
I would propose to add format provider fixed version of XmlExt.WriteAttribute into XmlWriterExt instead. This should fix the operational serialization issues as well as test procedures.

src/Spatial/Euclidean/Point2D.cs
```diff
@@ -361,8 +361,8 @@ void IXmlSerializable.ReadXml(XmlReader reader)
         /// <inheritdoc />
         void IXmlSerializable.WriteXml(XmlWriter writer)
         {
-            writer.WriteAttribute("X", this.X);
-            writer.WriteAttribute("Y", this.Y);
+            writer.WriteElement("X", this.X, "G15");
```
JohanLarsson, line 364: Changing from attribute to element is a breaking change. Elements would probably have been the better choice.

`"R"` is a better format specifier. https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#RFormatString

jkalias:
@rsinnig @JohanLarsson any progress on this?

JohanLarsson:
No progress

JohanLarsson:
I don't remember either, sorry.

jkalias:
Ok, fair enough.
I propose we give this issue one month time, and if it is not addressed till then, we reject it. Agree? @rsinnig @JohanLarsson

jkalias:
I think I understand now what the issue is, having worked in #208. I also think that `WriteElement` is a better approach.

PR #182, opened 2020-09-11
Update Vector3DHomogeneous.cs

osbordh:
Removed links that are dead from field comments.
Fixed constructor that take Vector argument (W component was incorrect).  Also replaced :this(x,y,z,w) with assignments in the constructor body, since you must check vector length before calling this(x,y,z,w).
Fixed ToVector3D().  Copy/paste error  (all components were this.X).

### DIFF
--- a/src/Spatial/Projective/Vector3DHomogeneous.cs
+++ b/src/Spatial/Projective/Vector3DHomogeneous.cs
@@ -12,22 +12,22 @@ namespace MathNet.Spatial.Projective
     internal struct Vector3DHomogeneous : IEquatable<Vector3DHomogeneous>
     {
         /// <summary>
-        /// Using public fields for performance. 
+        /// Using public fields for performance.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Spatial/Projective/Vector3DHomogeneous.cs
```diff
@@ -51,12 +51,15 @@ public Vector3DHomogeneous(double x, double y, double z, double w)
         /// </summary>
         /// <param name="vector">A mathnet.numerics vector</param>
         private Vector3DHomogeneous(Vector<double> vector)
-            : this(vector.At(0), vector.At(1), vector.At(2), vector.At(2))
         {
             if (vector.Count != 4)
             {
                 throw new ArgumentException("Size must be 4");
             }
+            this.X = vector.At(0);
```
jkalias, line 59: Please integrate master into this branch, and remove all ```this.``` qualifiers from the file.

PR #186, opened 2021-04-27
Add sphere3D

sangminpakr:


src/Spatial/Euclidean/Sphere3D.cs
```diff
@@ -0,0 +1,245 @@
...
+        /// <param name="radius">The radius of the Sphere</param>
+        public Sphere3D(Point3D centerPoint, double radius)
+        {
+            // Anypoint is same to other point.
+            if (radius <= 0)
+            {
+                throw new ArgumentException("The radius is negative.");
+            }
+
+            this.CenterPoint = centerPoint;
```
jkalias, line 37: Please remove all ```this.``` qualifiers

src/Spatial/Euclidean/Sphere3D.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        /// <summary>
+        /// Creates an instance of <see cref="Sphere3D"/>sphere which lie along its circumference of tetrahydron made of four points.
+        /// </summary>
+        /// <param name="p1">The first point on the sphere</param>
+        /// <param name="p2">The second point on the sphere</param>
+        /// <param name="p3">The third point on the sphere</param>
+        /// <param name="p4">The last point on the sphere</param>
+        public static Sphere3D FromFourPoints(Point3D p1, Point3D p2, Point3D p3, Point3D p4)
+        {
```
jkalias, line 95: I would favor an implementation which is easier to follow, based for example on intersections of planes and lines

src/Spatial/Euclidean/Sphere3D.cs
jkalias, line 1: I think it would be useful to have additional methods, like
* intersection of sphere with line/ray
* intersection of sphere with plane
* intersection of sphere with another sphere
* functions which check if a point lies inside/on/outside of the circle

src/Spatial.Tests/Euclidean/Sphere3DTests.cs
```diff
@@ -0,0 +1,84 @@
+﻿
+using System;
+using MathNet.Spatial.Euclidean;
+using NUnit.Framework;
+
+namespace MathNet.Spatial.UnitTests.Euclidean
```
jkalias, line 6: Please fix namespace

src/Spatial/Euclidean/Sphere3D.cs
JohanLarsson: IIRC we use StyleCop with default settings in this repro so if so `this.` is correct.

src/Spatial/Euclidean/Sphere3D.cs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        /// <inheritdoc />
+        [Pure]
+        public override bool Equals(object obj) => obj is Sphere3D c && this.Equals(c);
+
+        /// <inheritdoc />
+        [Pure]
+        public override int GetHashCode() => HashCode.Combine(this.CenterPoint, this.Radius);
+    }
```
JohanLarsson, line 239: nit: empty rows here

src/Spatial/Euclidean/Sphere3D.cs
jkalias: I have already removed all "this." references in a previous [commit](https://github.com/mathnet/mathnet-spatial/commit/e3d3c8c469cb4fb639296b469663f266972e0328), and the CI build didn't complain, so I thought it was an acceptable change

jkalias:
@fcsangminpark any updates on this?

Issue #178, opened 2020-06-05
LineSegment2D - TryIntersect - should this return the intersection of lines or line segments?


gtodd258:
When attempting to find the intersection between two _line segments_, the results are what I would expect to see when trying to find the intersection between two _lines_. My understanding is that line segments, as opposed to lines, do not continue infinitely in either direction, but the TryIntersect method returns results assuming that they do.

For example, from IntersectWithTest in LineSegment2DTests.cs, the first TestCase points as _line segments_ do **not** intersect, but if they were Line2D then they would intersect. See:

![2020-06-04 22_29_04-Book1 - Excel](https://user-images.githubusercontent.com/28796178/83840266-f4d1c880-a6b2-11ea-8137-56f9fee366af.png)

I'm attempting to use this in a manner where I only want the TryIntersect to return a valid intersection as the passed by reference Point2D if the _segments_ themselves intersect, which is what I would expect based on the naming of the LineSegment2D class and TryIntersect method.

If this is the intended behavior then I have a workaround, otherwise I'd be willing to help improve the project.

JohanLarsson:
Looks like you found a bug, my expectation is the same as yours given that we have the two types. This is a breaking change but one we should probably make.

gtodd258:
Taking another look at this, it does appear that the correct boolean value is being returned, but I am unclear as to what the desired functionality with the intersection Point2D parameter is. The intersection as if the line segments were treated as lines is returned so long as an intersection is possible.

The parameter comment reads:
> When this method returns, contains the intersection point, if the conversion succeeded, or the default point if the conversion failed.

so I suppose it depends on what exactly is meant by the conversion succeeding or failing.

This may be a non-issue, but the intersection Point2D parameter lead to some confusion on my end.

JohanLarsson:
That piece of documentation reads like a copy-paste accident.

Issue #111, opened 2017-12-10
Null return consistency


AQuentin2033:
Proposal to prefer Try methods for cases where the action is not a guarantee.  
For example TryIntersect is preferable over Intersect returning Nullable

PR #187, opened 2021-04-29
Fix Line2D.TryIntersect to pass nullable Point2D

f-frhs:
This pull-request is asking for help to solve #178.

The problem is that `LineSegment2D.TryIntersect()` always passes by reference `Point2D` of intersection, regardless of whether the two `LineSegment2D` actually have a valid intersection or not.

In my fix, `LineSegment2D.TryIntersect()` passes by ref `Point2D?`, which can indicate that there is no intersection.
My fix used #111 as reference.

I'm not sure that this breaking change is acceptable or not.
If there is another way to solve this, I'm happy to learn it.

I considered these fix-plans:
1. Leave it as it is
2. Pass by ref default(Point2D)
3. Pass by ref Nullable&lt;Point2D&gt;
4. Pass by ref Point2D.Null (kinda)
5. Return Nullable&lt;Point2D&gt; instead of bool


src/Spatial/Euclidean/LineSegment2D.cs
```diff
@@ -134,19 +134,19 @@ public Point2D ClosestPointTo(Point2D p)
...
-        /// <param name="intersection">When this method returns, contains the intersection point, if the conversion succeeded, or the default point if the conversion failed.</param>
-        /// <param name="tolerance">The tolerance used when checking if the lines are parallel</param>
+        /// <param name="other">The other line segment to compute the intersection with</param>
+        /// <param name="intersection">The intersection if it exists; otherwise null</param>
+        /// <param name="tolerance">The tolerance used when checking if the line segments are parallel</param>
         /// <returns>True if an intersection exists; otherwise false</returns>
         [Pure]
-        public bool TryIntersect(LineSegment2D other, out Point2D intersection, Angle tolerance)
+        public bool TryIntersect(LineSegment2D other, out Point2D? intersection, Angle tolerance)
         {
```
jkalias, line 146: Initialization ```intersection = null;```

src/Spatial/Euclidean/LineSegment2D.cs
```diff
@@ -134,19 +134,19 @@ public Point2D ClosestPointTo(Point2D p)
...
             if (this.IsParallelTo(other, tolerance))
             {
-                intersection = default(Point2D);
+                intersection = null;
```
jkalias, line 149: Not needed here if the initialization is done at the function entry point

src/Spatial/Euclidean/LineSegment2D.cs
```diff
@@ -159,8 +159,17 @@ public bool TryIntersect(LineSegment2D other, out Point2D intersection, Angle to
...
 
-            intersection = p + (t * r);
-            return (0.0 <= t && t <= 1.0) && (0.0 <= u && u <= 1.0);
+            var isIntersected = (0.0 <= t && t <= 1.0) && (0.0 <= u && u <= 1.0);
+            if (isIntersected)
+            {
+                intersection = p + (t * r);
+                return true;
+            }
+            else
```
jkalias, line 168: You can simplify all this by simply ```return intersection != null;``` and completely removing the ```else``` case.

jkalias (changes requested):

jkalias (approved):

src/Spatial/Euclidean/LineSegment2D.cs
f-frhs: Thanks for a comment.

How about the following case?
83840266-f4d1c880-a6b2-11ea-8137-56f9fee366af

This comes from #178.

f-frhs: Thanks for a comment.
Your understanding is correct that we can verify it by using `LineSegment2DTests.IntersectWithTest2`.

I'm happy with the merged code. Thanks.

jkalias: Thank YOU for opening the issue and providing a PR 😃

PR #191, opened 2021-08-19
Increase PolyLine performance when indexing Vertices

bradtglass:
Expose `IReadOnlyList<T>` instead of `IEnumerable<T>` for `PolyLine.Vertices`

This change allows external methods to index or operate on the vertices
without consuming additional memory by using `ToList()` or `ToArray()` or enumerating the entire collection
multiple times. This change still ensures that the internal list is immutable.

I don't think this is a breaking changes because `IReadOnlyList<T>` implements `IEnumerable<T>`, so anything that relies on `IEnumerable<T>` will still work.

This is implemented by using `ReadOnlyCollection<T>` internally instead of `List<T>`.

I think the same change would be appropriate for `Polygon2D` too but I'm not sure if there was a specific reason for using a custom implementation of `ImmutableList<T>` instead of `ReadOnlyCollection<T>`.

### DIFF
--- a/src/Spatial/Euclidean/PolyLine2D.cs
+++ b/src/Spatial/Euclidean/PolyLine2D.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using HashCode = MathNet.Spatial.Internals.HashCode;
@@ -15,7 +16,7 @@ namespace MathNet.Spatial.Euclidean
         /// <summary>
         /// Internal storage for the points
         /// </summary>
-        private readonly List<Point2D> points;
+        private readonly ReadOnlyCollection<Point2D> points;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PolyLine2D"/> class.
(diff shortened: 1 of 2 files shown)


cdrnet:
Thank you!

Looks like I first need to upgrade the SDK to fix CI build before I can merge.

src/Spatial/Euclidean/PolyLine2D.cs
```diff
@@ -15,7 +16,7 @@ public class PolyLine2D : IEquatable<PolyLine2D>
         /// <summary>
         /// Internal storage for the points
         /// </summary>
-        private readonly List<Point2D> points;
+        private readonly ReadOnlyCollection<Point2D> points;
```
jkalias, line 19: I am in a trajectory where I try to remove all redundant ```this.``` qualifiers from all classes. So please rename this private field to ```_points``` and remove all ```this.``` from these files.

src/Spatial/Euclidean/PolyLine3D.cs
```diff
@@ -14,7 +15,7 @@ public class PolyLine3D : IEquatable<PolyLine3D>
         /// <summary>
         /// An internal list of points
         /// </summary>
-        private readonly List<Point3D> points;
+        private readonly ReadOnlyCollection<Point3D> points;
```
jkalias, line 18: Same comment as in ```PolyLine2D.cs```

jkalias (changes requested):
Please integrate master into this master. The CI bug has been fixed, so we can check if this PR passes

jkalias (approved):

PR #194, opened 2022-04-04
Fix order of YPR transformation

### LINKED ISSUE #180: Rotation(yaw, pitch, roll) uses extrinsic rotations
The method `CoordinateSystem.Rotation(Angle yaw, Angle pitch, Angle roll)` uses extrinsic rotations instead of intrinsic rotations as defined on [Wikipedia](https://en.wikipedia.org/wiki/Euler_angles#Proper_Euler_angles).

jakehedlund:
See https://en.wikipedia.org/wiki/Rotation_matrix#General_rotations for correct order of application. 

Counterintuitively, the YPR rotations must be applied in reverse order to be correct. The resultant rotation matrix should be Rz(Ry(Rx)) instead of Rx(Ry(Rz)) where z is yaw, y is pitch, and x is roll. This is the correct intrinsic spatial rotation as shown in the Wikipedia link above. 

Fixes #180.

For easy verification, an excel file to compute the correct rotation matrix is attached. 
[rotationmatrix.xlsx](https://github.com/mathnet/mathnet-spatial/files/8412609/rotationmatrix.xlsx)

### DIFF
--- a/src/Spatial/Euclidean/CoordinateSystem.cs
+++ b/src/Spatial/Euclidean/CoordinateSystem.cs
@@ -258,11 +258,10 @@ namespace MathNet.Spatial.Euclidean
         /// <returns>A rotated coordinate system</returns>
         public static CoordinateSystem Rotation(Angle yaw, Angle pitch, Angle roll)
         {
-            var cs = new CoordinateSystem();
             var yt = Yaw(yaw);
             var pt = Pitch(pitch);
             var rt = Roll(roll);
-            return rt.Transform(pt.Transform(yt.Transform(cs)));
+            return yt.Transform(pt.Transform(rt));
         }
 
         /// <summary>


cdrnet:
Thank you!

Looks like I first need to upgrade the SDK to fix CI build before I can merge.

jkalias:
SDK has been updated to 5.0.200, so the CI build runs through.

src/Spatial/Euclidean/CoordinateSystem.cs
```diff
@@ -258,13 +258,44 @@ public static CoordinateSystem Rotation(Angle angle, Vector3D v)
         /// <returns>A rotated coordinate system</returns>
         public static CoordinateSystem Rotation(Angle yaw, Angle pitch, Angle roll)
```
jkalias, line 259: There is no need to keep the slow version around, now that we have a faster version which returns the same result. Please remove it.

src/Spatial/Euclidean/CoordinateSystem.cs
```diff
@@ -258,13 +258,44 @@ public static CoordinateSystem Rotation(Angle angle, Vector3D v)
...
-            var cs = new CoordinateSystem();
             var yt = Yaw(yaw);
             var pt = Pitch(pitch);
             var rt = Roll(roll);
-            return rt.Transform(pt.Transform(yt.Transform(cs)));
+            return yt.Transform(pt.Transform(rt));
         }
 
+        /// <summary>
+        /// Create a rotated CoordinateSystem in one calculation instead of 3 separate rotations.
```
jkalias, line 268: I would add a URL link to the wiki page section which documents this calculation

src/Spatial.Tests/Euclidean/CoordinateSystemTest.cs
```diff
@@ -91,15 +97,43 @@ public void RotationYawPitchRoll(string yaws, string pitchs, string rolls, strin
...
             foreach (var coordinateSystem in coordinateSystems)
             {
                 var rotatedPoint = coordinateSystem.Transform(p);
-                AssertGeometry.AreEqual(expected, rotatedPoint);
+                AssertGeometry.AreEqual(expected, rotatedPoint, 1e-3);
             }
         }
 
+        [Test]
+        public void RotationSpeedTest()
```
jkalias, line 111: We don't need these speed tests once the legacy (slow) version is removed

src/Spatial.Tests/Euclidean/CoordinateSystemTest.cs
```diff
@@ -82,6 +82,12 @@ public void RotationAroundVector(string ps, string @as, string vs, string eps)
         [TestCase("0°", "-90°", "0°", "1, 2, 3", "-3, 2, 1")]
         [TestCase("0°", "0°", "90°", "1, 2, 3", "1, -3, 2")]
         [TestCase("0°", "0°", "-90°", "1, 2, 3", "1, 3, -2")]
+        [TestCase("90°", "90°", "90°", "1, 2, 3", "3, 2, -1")]
+        [TestCase("90°", "0°", "90°", "1, 2, 3", "3, 1, 2")]
+        [TestCase("180°", "0°", "90°", "1, 2, 3", "-1, 3, 2")]
+        [TestCase("180°", "-90°", "0°", "1, 2, 3", "3, -2, 1")]
+        [TestCase("90°", "10°", "0°", "1, 2, 3", "-2, 1.506, 2.781")]
+        [TestCase("90°", "10°", "30°", "1, 2, 3", "-0.232, 1.609, 3.370")]
```
jkalias, line 90: I would another case, for which no angle is 90° or 180°, just for the sake of completeness (eg. 15°, -23°, 48°).

jkalias (changes requested):

jkalias:
Please merge the master branch in your branch, so that you also gain from the passing CI build. Otherwise we cannot merge this.

PR #216, opened 2023-04-05
Best plane fit from points

f-frhs:
I added `Plane.CreateFittedPlaneFrom(points)` method with some unit tests for it.

I would like to discuss the following topic:
- [ ] Were testcases I added sufficient?
- implementation
  - [x] which is nice to evaluate?
    a. (z)axis-aligned distance
    b. orthogonal distance

### DIFF
--- a/src/Spatial.Tests/Euclidean/PlaneTest.cs
+++ b/src/Spatial.Tests/Euclidean/PlaneTest.cs
@@ -85,6 +85,31 @@ namespace MathNet.Spatial.Tests.Euclidean
             AssertGeometry.AreEqual(expected, projectedPoint, eps);
         }
 
+        [Test, Sequential]
+        public void BestFit_From3RandomPoints(
+            [Random(0d, 10d, 3)] double p0x,
+            [Random(0d, 10d, 3)] double p0y,
+            [Random(0d, 10d, 3)] double p0z,
+            [Random(0d, 10d, 3)] double p1x,
+            [Random(0d, 10d, 3)] double p1y,
+            [Random(0d, 10d, 3)] double p1z,
+            [Random(0d, 10d, 3)] double p2x,
+            [Random(0d, 10d, 3)] double p2y,
+            [Random(0d, 10d, 3)] double p2z
+        )
+        {
+            var ps = new[]
+            {
+                new Point3D(p0x, p0y, p0z),
+                new Point3D(p1x, p1y, p1z),
+                new Point3D(p2x, p2y, p2z),
+            };
+            var actual = Plane.CreateFittedPlaneFrom(ps);
+
+            var expected = Plane.FromPoints(ps[0], ps[1], ps[2]);
+            AssertGeometry.AreEqual(expected, actual);
+        }
+
         [TestCase(ZeroPoint, X, Y)] //all 3 points are on the plane z=0. normal vector is +ez.
         [TestCase(ZeroPoint, Y, Z)] //all 3 points are on the plane x=0. normal vector is +ex.
         [TestCase(ZeroPoint, Z, X)] //all 3 points are on the plane y=0. normal vector is +ey.
--- a/src/Spatial.Tests/Euclidean/PlaneTest.cs
+++ b/src/Spatial.Tests/Euclidean/PlaneTest.cs
@@ -104,7 +104,7 @@ namespace MathNet.Spatial.Tests.Euclidean
                 new Point3D(p1x, p1y, p1z),
                 new Point3D(p2x, p2y, p2z),
             };
-            var actual = Plane.CreateFittedPlaneFrom(ps);
+            var actual = Plane.BestFit(ps);
 
             var expected = Plane.FromPoints(ps[0], ps[1], ps[2]);
             AssertGeometry.AreEqual(expected, actual);
@@ -121,7 +121,7 @@ namespace MathNet.Spatial.Tests.Euclidean
                 Point3D.Parse(sP2),
                 Point3D.Parse(sP3),
             }.ToArray();
-            var aPlane = Plane.CreateFittedPlaneFrom(ps);
+            var aPlane = Plane.BestFit(ps);
 
             var tolerance = 1e-16;
             Assert.That(aPlane.SignedDistanceTo(ps[0]), Is.EqualTo(0).Within(tolerance), "ps[0]");
@@ -137,7 +137,7 @@ namespace MathNet.Spatial.Tests.Euclidean
             //https://math.stackexchange.com/questions/99299/best-fitting-plane-given-a-set-of-points/1652383#1652383
 
             var points = sPoints.Split(';').Select(s => Point3D.Parse(s));
-            var actual = Plane.CreateFittedPlaneFrom(points);
+            var actual = Plane.BestFit(points);
 
             //normalize (eA, eB, eC and eD).
             //both 2 following equations mean the same plane.
--- a/src/Spatial.Tests/Euclidean/PlaneTest.cs
+++ b/src/Spatial.Tests/Euclidean/PlaneTest.cs
@@ -122,10 +122,6 @@ namespace MathNet.Spatial.Tests.Euclidean
             var eNormalizedOffset = eOffset / normal.Length;
             var expected = new Plane(normal.Negate().Normalize(), eNormalizedOffset);
 
-            //for debugging purpose
-            TestContext.WriteLine($"{nameof(actual)}={actual}");
-            TestContext.WriteLine($"{nameof(expected)}={expected}");
-
             var tolerance = 6e-3; // for this example
             AssertGeometry.AreEqual(expected, actual);
         }
--- a/src/Spatial.Tests/Euclidean/PlaneTest.cs
+++ b/src/Spatial.Tests/Euclidean/PlaneTest.cs
@@ -103,31 +103,7 @@ namespace MathNet.Spatial.Tests.Euclidean
             Assert.That(aPlane.SignedDistanceTo(ps[1]), Is.EqualTo(0).Within(tolerance), "ps[1]");
             Assert.That(aPlane.SignedDistanceTo(ps[2]), Is.EqualTo(0).Within(tolerance), "ps[2]");
         }
-
-        [TestCase(ZeroPoint, X, Y, XY, Z, "0,0,+1")] //all 4 points fall on the plane z=+1
-        [TestCase(ZeroPoint, X, Y, XY, Z, "0,0,-1")] //all 4 points fall on the plane z=-1
-        [TestCase(ZeroPoint, Y, Z, YZ, X, "+1,0,0")] //all 4 points fall on the plane x=+1
-        [TestCase(ZeroPoint, Y, Z, YZ, X, "-1,0,0")] //all 4 points fall on the plane x=-1
-        [TestCase(ZeroPoint, Z, X, ZX, Y, "0,+1,0")] //all 4 points fall on the plane y=+1
-        [TestCase(ZeroPoint, Z, X, ZX, Y, "0,-1,0")] //all 4 points fall on the plane y=-1
-        public void CreateFittedPlaneFrom_With4PointsOnAxisAlignedPlane(string sV1, string sV2, string sV3, string sV4, string sENormal, string sRootPoint)
-        {
-            var rootPoint = Point3D.Parse(sRootPoint);
-            var ps = new[]
-            {
-                Vector3D.Parse(sV1),
-                Vector3D.Parse(sV2),
-                Vector3D.Parse(sV3),
-                Vector3D.Parse(sV4),
-            }.Select(deltaV => rootPoint + deltaV);
-            var actual = Plane.CreateFittedPlaneFrom(ps);
-
-            var eNormal = UnitVector3D.Parse(sENormal);
-            var eRootPoint = Point3D.Parse(sRootPoint);
-            var expected = new Plane(eNormal, eRootPoint);
-            Assert.That(actual, Is.EqualTo(expected));
-        }
-
+        
         [TestCase("1,1,9;1,2,14;1,3,20;2,1,11;2,2,17;2,3,23;3,1,15;3,2,20;3,3,26", -21.7240278973, -41.0640090729, 7.3269978417, 1)]
         public void CreateFittedPlane(string sPoints, double eA, double eB, double eC, double eOffset)
         {
--- a/src/Spatial.Tests/Euclidean/PlaneTest.cs
+++ b/src/Spatial.Tests/Euclidean/PlaneTest.cs
@@ -1,9 +1,10 @@
-﻿// ReSharper disable InconsistentNaming
+// ReSharper disable InconsistentNaming
 
-using System;
-using System.Linq;
 using MathNet.Spatial.Euclidean;
 using NUnit.Framework;
+using System;
+using System.Linq;
+using MathNet.Numerics.LinearAlgebra;
 
 namespace MathNet.Spatial.Tests.Euclidean
 {
@@ -11,7 +12,11 @@ namespace MathNet.Spatial.Tests.Euclidean
     public class PlaneTest
     {
         private const string X = "1; 0 ; 0";
+        private const string Y = "0; 1; 0";
         private const string Z = "0; 0; 1";
+        private const string XY = "1; 1; 0";
+        private const string YZ = "0; 1; 1";
+        private const string ZX = "1; 0; 1";
         private const string NegativeZ = "0; 0; -1";
         private const string ZeroPoint = "0; 0; 0";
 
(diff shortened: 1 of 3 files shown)


src/Spatial/Euclidean/Plane.cs
```diff
@@ -74,6 +77,30 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Plane"/> struct.
+        /// Constructs a Plane from the given points.
+        /// </summary>
+        /// <param name="points">The points</param>
+        /// <returns>Fitted <see cref="Plane"/></returns>
+        /// <remarks>this method uses SVD to fit the plane.</remarks>
+        public static Plane CreateFittedPlaneFrom(IEnumerable<Point3D> points)
+        {
+            var c = Point3D.Centroid(points);
```
jkalias, line 89: I would use more descriptive variable names, like `throughPoint` instead of just `c`.

src/Spatial/Euclidean/Plane.cs
```diff
@@ -74,6 +77,30 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+
+            var A = CreateMatrix.DenseOfRowVectors(
```
jkalias, line 91: This is also called the [scatter matrix](https://math.stackexchange.com/questions/3501135/fitting-a-plane-to-points-using-svd)

src/Spatial/Euclidean/Plane.cs
```diff
@@ -74,6 +77,30 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+                points.Select(p => p - c)
+                    .Select(p => CreateVector.DenseOfArray(new double[] { p.X, p.Y, p.Z })));
```
jkalias, line 93: What I would do in this case, is to first allocate the necessary matrix space (`new DenseMatrix(rows, columns)`), and then write a single for-each to populate each row of it. This way we could easily parallelize the loop in case we need it, if for example we have a LOT of points.

src/Spatial/Euclidean/Plane.cs
```diff
@@ -74,6 +77,30 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+
+            var svd = A.Svd(true);
+            var matV = svd.VT.Transpose();
+            var theIndex = svd.S.Count-1; // in this case, theIndex = 2.
```
jkalias, line 97: I think you can entirely skip this variable and use `var normal = UnitVector3D.OfVector(matV.Column(matV.Cols - 1));`. This better captures semantically the intention, that we need the last eigenvector of V.

src/Spatial/Euclidean/Plane.cs
```diff
@@ -74,6 +77,30 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
+            var normal = UnitVector3D.OfVector(matV.Column(theIndex));
+
+            var result = new Plane(normal, c);
```
jkalias, line 100: `var bestFit = new Plane(normal, c);`

src/Spatial/Euclidean/Plane.cs
```diff
@@ -74,6 +77,30 @@ public Plane(Point3D rootPoint, UnitVector3D normal)
...
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Plane"/> struct.
+        /// Constructs a Plane from the given points.
+        /// </summary>
+        /// <param name="points">The points</param>
+        /// <returns>Fitted <see cref="Plane"/></returns>
+        /// <remarks>this method uses SVD to fit the plane.</remarks>
+        public static Plane CreateFittedPlaneFrom(IEnumerable<Point3D> points)
```
jkalias, line 87: I think we need to locally transform the `IEnumerable<Point3D>` to a `List<Point3D>`. Then we wouldn't traverse the collection multiple times.

I also think that a better name would just be `BestFit`

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
             AssertGeometry.AreEqual(expected, projectedPoint, eps);
         }
 
+        [TestCase(ZeroPoint, X, Y)] //all 3 points are on the plane z=0. normal vector is +ez.
+        [TestCase(ZeroPoint, Y, Z)] //all 3 points are on the plane x=0. normal vector is +ex.
+        [TestCase(ZeroPoint, Z, X)] //all 3 points are on the plane y=0. normal vector is +ey.
+        public void CreateFittedPlaneFrom_PointsShouldBeOnTheFittedPlane(string sP1, string sP2, string sP3)
```
jkalias, line 91: I would also add another test case, with 3 general points, not just the ones which lie on the x, y and z axes.

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
...
+            var ps = new[]
+            {
+                Point3D.Parse(sP1),
+                Point3D.Parse(sP2),
+                Point3D.Parse(sP3),
+            }.ToArray();
+            var aPlane = Plane.CreateFittedPlaneFrom(ps);
+
+            var tolerance = 1e-16;
+            Assert.That(aPlane.SignedDistanceTo(ps[0]), Is.EqualTo(0).Within(tolerance), "ps[0]");
```
jkalias, line 102: `AssertGeometry.AreEqual(expectedPlane, actualPlane);`

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
...
+            Assert.That(aPlane.SignedDistanceTo(ps[2]), Is.EqualTo(0).Within(tolerance), "ps[2]");
+        }
+
+        [TestCase(ZeroPoint, X, Y, XY, Z, "0,0,+1")] //all 4 points fall on the plane z=+1
+        [TestCase(ZeroPoint, X, Y, XY, Z, "0,0,-1")] //all 4 points fall on the plane z=-1
+        [TestCase(ZeroPoint, Y, Z, YZ, X, "+1,0,0")] //all 4 points fall on the plane x=+1
+        [TestCase(ZeroPoint, Y, Z, YZ, X, "-1,0,0")] //all 4 points fall on the plane x=-1
+        [TestCase(ZeroPoint, Z, X, ZX, Y, "0,+1,0")] //all 4 points fall on the plane y=+1
+        [TestCase(ZeroPoint, Z, X, ZX, Y, "0,-1,0")] //all 4 points fall on the plane y=-1
+        public void CreateFittedPlaneFrom_With4PointsOnAxisAlignedPlane(string sV1, string sV2, string sV3, string sV4, string sENormal, string sRootPoint)
```
jkalias, line 113: It's not really clear to me, what this test is supposed to guarantee.

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
...
+            }.Select(deltaV => rootPoint + deltaV);
+            var actual = Plane.CreateFittedPlaneFrom(ps);
+
+            var eNormal = UnitVector3D.Parse(sENormal);
+            var eRootPoint = Point3D.Parse(sRootPoint);
+            var expected = new Plane(eNormal, eRootPoint);
+            Assert.That(actual, Is.EqualTo(expected));
+        }
+
+        [TestCase("1,1,9;1,2,14;1,3,20;2,1,11;2,2,17;2,3,23;3,1,15;3,2,20;3,3,26", -21.7240278973, -41.0640090729, 7.3269978417, 1)]
```
jkalias, line 131: We need more tests, which cover what happens when we have less than 3 points (2, 1, or none).

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
...
+
+            //for debugging purpose
+            TestContext.WriteLine($"{nameof(actual)}={actual}");
+            TestContext.WriteLine($"{nameof(expected)}={expected}");
+
+            var tolerance = 6e-3; // for this example
+            Assert.That(actual.A, Is.EqualTo(expected.A).Within(tolerance), "A");
+            Assert.That(actual.B, Is.EqualTo(expected.B).Within(tolerance), "B");
+            Assert.That(actual.C, Is.EqualTo(expected.C).Within(tolerance), "C");
+            Assert.That(actual.D, Is.EqualTo(expected.D).Within(tolerance), "D");
```
jkalias, line 156: `AssertGeometry.AreEqual(expectedPlane, actualPlane);`

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
...
+            // eq2:-Ax - By - Cz - D = 0
+            var normal = new Vector3D(eA, eB, eC);
+            var eNormalizedOffset = eOffset / normal.Length;
+            var expected = new Plane(normal.Negate().Normalize(), eNormalizedOffset);
+
+            //for debugging purpose
+            TestContext.WriteLine($"{nameof(actual)}={actual}");
+            TestContext.WriteLine($"{nameof(expected)}={expected}");
+
+            var tolerance = 6e-3; // for this example
```
jkalias, line 152: This seems like a magic number. Perhaps the values from the original example are not as exact as we want them to be. I would adjust the values in our example, so that the default tolerance would pass the test.

src/Spatial.Tests/Euclidean/PlaneTest.cs
```diff
@@ -80,6 +85,77 @@ public void ProjectPointOn(string ps, string rootPoint, string unitVector, strin
...
+
+            //normalize (eA, eB, eC and eD).
+            //both 2 following equations mean the same plane.
+            // eq1: Ax + By + Cz + D = 0
+            // eq2:-Ax - By - Cz - D = 0
+            var normal = new Vector3D(eA, eB, eC);
+            var eNormalizedOffset = eOffset / normal.Length;
+            var expected = new Plane(normal.Negate().Normalize(), eNormalizedOffset);
+
+            //for debugging purpose
```
jkalias, line 148: We don't need debugging for tests

jkalias (changes requested):
Thank you for this PR, it seems interesting. Let me know if my comments are not clear enough.

jkalias:
>       * [ ]  which is nice to evaluate?
>         a. (z)axis-aligned distance
>         b. orthogonal distance

I think your implementation is correct, we should evaluate the best fit based on the orthogonal euclidean distance.

src/Spatial/Euclidean/Plane.cs
f-frhs: will do, thanks.

src/Spatial/Euclidean/Plane.cs
f-frhs: If I understood your comment correctly, it seems to me that the [scatter matrix](https://math.stackexchange.com/questions/3501135/fitting-a-plane-to-points-using-svd) means a different matrix from the matrix `A` written in the code.

To say that these 2 matrices are different, I checked their sizes(=rows count and columns count).
The [scatter matrix](https://math.stackexchange.com/questions/3501135/fitting-a-plane-to-points-using-svd) is an 3x3 matrix, 
and the matrix `A` is an nx3 matrix,
where $n$ is the count of points.

I agree with you to give variable a descriptive name.
Do you have any ideas?

src/Spatial/Euclidean/Plane.cs
f-frhs: will do, thanks.
I think that the parallelization is out of scope for me.
Thanks again for your point of view.

f-frhs: Does this fix follow the policy of Math.NET Spatial?
I'm afraid the case where you maintainers respect RAII([Resource acquisition is initialization](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization)) principle and I would break it.

src/Spatial/Euclidean/Plane.cs
f-frhs: I'm afraid that I didn't define any properties the *best* Fit plane should have.
Can I use `fittedPlane` for this?

src/Spatial.Tests/Euclidean/PlaneTest.cs
f-frhs: I totally agree with you.
I think I should remove this test.

src/Spatial.Tests/Euclidean/PlaneTest.cs
f-frhs: `Plane.CreateFittedPlaneFrom(points)` is not for the testcases with insufficient points.
`Plane.CreateFittedPlaneFrom(points)` is for over-determined system.
Instead, `Plane.Validate(points)` will do for this purpose.

It goes a detailed implementation, `Plane.Validate(points)` will be something like:
```csharp
public static void Validate(IEnumerable<Point3D> points)
{
    //      [p0x, p0y, p0z]
    //      [p1x, p1y, p1z]
    // mat= [p2x, p2y, p2z]
    //      [p3x, p3y, p3z]
    //      [p4x, p4y, p4z]
    //      [..., ..., ...]

    //this expression will be fixed, according to the discussion
    var mat = CreateMatrix.DenseOfRowVectors(
                points.Select(p => CreateVector.DenseOfArray(new double[] { p.X, p.Y, p.Z })));

    var rankOfMat = mat.Rank();  // the library uses SVD.
    if(rankOfMat <= 0) // DoF=0
    {
        throws new Exception("The same point?");
    }
    if(rankOfMat <= 1) // DoF=1
    {
        throws new Exception("Are they collinear?");
    }

    // do nothing intentionally
    // because there are no problem, we can span 2d-vector-space with given points
}
```



If few (2 or less) points are given, the method `Plane.TryFromPonits()` should return `false`. It is similar to `int.TryParse(string)`'s behavior.

src/Spatial.Tests/Euclidean/PlaneTest.cs
f-frhs: True. I wish I could have.
I need time to adjust `AssertGeometry.AreEqual(PlaneA, PlaneB)` to pass this test
where
```csharp
PlaneA.Normal == -PlaneB.Normal;
PlaneA.RootPoint == PlaneB.RootPoint;
```

src/Spatial.Tests/Euclidean/PlaneTest.cs
f-frhs: ok. thank you for your cooperation.

src/Spatial.Tests/Euclidean/PlaneTest.cs
f-frhs: thanks for a comment.
I'll remove them.

src/Spatial/Euclidean/Plane.cs
f-frhs: > we should evaluate the best fit based on the orthogonal euclidean distance.

Sorry, I couldn't read all your comments at that moment. my fault.
You had already defined "the best fit" for this problem.
Now, we can use `var bestFit = new Plane(normal, c);`

src/Spatial/Euclidean/Plane.cs
jkalias: My apologies, you are right. The scatter matrix (`M` in the paper) would be `A^T * A` in your notation. Perhaps we can call this `relativePointMatrix`, since this is exactly what it is.

src/Spatial/Euclidean/Plane.cs
jkalias: > I think that the parallelization is out of scope for me.

Sure, I wasn't expecting this to be part of this PR. I don't want to prematurely optimize things in any case.

jkalias: > Does this fix follow the policy of Math.NET Spatial?
> I'm afraid the case where you maintainers respect RAII([Resource acquisition is initialization](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization)) policy and I would break it.

I am not sure I understand your question/concern here, can you explain it again please?

src/Spatial.Tests/Euclidean/PlaneTest.cs
jkalias: I think this is over-engineering the problem. We don't need a `Plane.Validate` method, we only need to check in this "best fit method" that before we dive in and calculate the SVD of A, its row count is more than or equal to 3. Something like `if (A.Rows < 3) { throw ...}` (the syntax might not be 100% perfect)

src/Spatial.Tests/Euclidean/PlaneTest.cs
jkalias: This is another problem I spotted in the `Plane` implementation after I took over the maintenance of this repo; the plane is not defined in its [Hesse normal form](https://en.wikipedia.org/wiki/Hesse_normal_form), which I find problematic.

Then the normal direction is uniquely defined and fixed, for any plane. In addition the equality operator should be adjusted, so that we don't check equality of the root points, but that something like `other.Conatins(rootPoint)`, where we check that the root point is contained in the other plane. There is nothing special in the choice of a root point for a plane, any point lying on it is equally valid.

I wanted to open another PR and fix this (breaking) change for the next (v0.8.0) release. Maybe you want to wait with this PR after I am finished (?)

jkalias (approved):

jkalias:
Fixed accuracy test in #218

Issue #197, opened 2023-02-07
CoordinateSystem: System.TypeInitializationException


kabircosta:
An exception occurs when I try to obtain the properties of a [CoordinateSystem](https://learn.microsoft.com/it-it/dotnet/api/system.typeinitializationexception?view=net-7.0) object, such as XAxis, YAxis, ZAxis and Origin , returning [System.TypeInitializationException](https://learn.microsoft.com/it-it/dotnet/api/system.typeinitializationexception?view=net-7.0).

jkalias:
Can you share some test code or how this can be reproduced?

PR #226, opened 2023-04-12
Breaking: Removed Matrix2D, renamed Matrix3D to RotationMatrix, and cleaned up CoordinateSystem

jkalias:


jkalias:
Related to #197

PR #228, opened 2023-04-13
Implemented IsParallelTo, IsPerpendicularTo, IsCollinearTo, IsCoplana…

kabircosta:
…rTo methods and Arc2D class. Tests are still pending.

### DIFF
--- a/src/Spatial/Euclidean/Plane.cs
+++ b/src/Spatial/Euclidean/Plane.cs
@@ -410,6 +410,66 @@ namespace MathNet.Spatial.Euclidean
             return ray.ThroughPoint + (t * ray.Direction);
         }
 
+        /// <summary> Checks if the plane is parallel to another plane. </summary>
+        /// <param name="other">The other plane.</param>
+        /// <param name="angleTolerance">The angle tolerance.</param>
+        /// <param name="distance">The distance between the two planes.</param>
+        /// <returns>True if the planes are parallel.</returns>
+        [Pure]
+        public bool IsParallelTo(
+            Plane other,
+            Angle angleTolerance = Precision.DoublePrecision * 2,
+            out double distance,
+            out bool sameDirection)
+        {
+            // We are parallel to ourselves
+            distance = double.NaN;
+
+            if (this == other)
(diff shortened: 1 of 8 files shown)


jkalias:
Thanks for your effort.

There is no `Line2D` and `Line3D` type anymore, just `Line` (a general 3D line). I am in a trajectory of eliminating all 2D/3D redundancies (according to #200 ) so the master branch is changing quite rapidly lately.

src/Spatial/Euclidean/Plane.cs
```diff
@@ -410,6 +410,66 @@ public Point3D IntersectionWith(Ray3D ray, double tolerance = float.Epsilon)
...
+
+        /// <summary>
+        /// Checks if the plane is coplanar to another plane.
+        /// </summary>
+        /// <param name="other">The plane to test against.</param>
+        /// <param name="angleTolerance">The angle tolerance to use for comparison.</param>
+        /// <param name="tolerance">The tolerance to use for comparison.</param>
+        /// <returns>True if the planes are coplanar, false otherwise.</returns>
+        [Pure]
+        public bool IsCoplanarTo(Plane other, Angle angleTolerance = Precision.DoublePrecision * 2, double tolerance = double.Epsilon)
```
jkalias, line 467: This is essentially the implementation of `Equals`, this is not needed

jkalias (changes requested):

src/Spatial/Euclidean/Plane.cs
kabircosta: Two planes can be coplanar, but in opposite directions.
For example:
Case 1) Coplanar and same direction: normal1 = normal2
Case 2) Coplanar and opposite direction: normal1 = normal2.Inverse

jkalias:
After the discussion in #229 and the reverted changes, we might have to revisit the comments here

jkalias:
Could you please split this PR into two distinct ones? The one containing only the additional functions, and another containing the new ```Arc2D``` class? It would then be easier to separate the concerns

PR #235, opened 2023-08-30
Add Vector2D.ToPoint2D() method with unit test

mymike00:
Similar method is present for Vector3D, so I added it to Vector2D too

### DIFF
--- a/src/Spatial/Euclidean/Vector2D.cs
+++ b/src/Spatial/Euclidean/Vector2D.cs
@@ -476,6 +476,15 @@ namespace MathNet.Spatial.Euclidean
         }
 
         /// <summary>
+        /// Returns a point equivalent to the vector
+        /// </summary>
+        /// <returns>A point</returns>
+        public Point2D ToPoint2D()
+        {
+            return new Point2D(this.X, this.Y);
+        }
+
+        /// <summary>
         /// Transforms a vector by multiplying it against a provided matrix
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Spatial/Euclidean/Vector2D.cs
```diff
@@ -475,6 +475,15 @@ public Vector2D Add(Vector2D v)
             return new Vector2D(X + v.X, Y + v.Y);
         }
 
+        /// <summary>
+        /// Returns a point equivalent to the vector
```
jkalias, line 479: The term _equivalent_ can be misleading. I would rather express this a bit more verbosely, like _Returns a point, whose position vector is equal to the current vector_

jkalias (changes requested):

src/Spatial/Euclidean/Vector2D.cs
mymike00: Should the [comment ](https://github.com/mathnet/mathnet-spatial/blob/f6079916f41912ac0dbfc59003bae673a30ae45f/src/Spatial/Euclidean/Vector3D.cs#L526)in Vector3D ToPoint3D() changed as well? I just copy-pasted mine from there...

jkalias: I would, yes, but it (sadly) seems that this project does not get any attention

jkalias (approved):

jkalias (approved):

PR #236, opened 2023-11-04
Bug in BaseChangeMatrix, TransformFromCoordSys, TransformToCoordSys

fairendil:
The bug occurred when calculating the BaseChangeMatrix and a copy of the coordinate system was rotated instead of the base coordinate system rotation. This resulted in the OffsetToBase being added twice, causing an issue in the transformation process.


jkalias:
Can you provide a regression unit test case for this bug fix please?

jkalias:
@fairendil is this still relevant?

fairendil:
No. It's still working wrong and we are using Transform(Point3D/Vector3D)

jkalias:
> No. It's still working wrong and we are using Transform(Point3D/Vector3D)

So it IS relevant if it’s still wrong, correct?
Would you mind including some regression unit tests for this case?

PR #237, opened 2023-11-25
Circle2D intersection with Line(Segment)2D

f-frhs:
I implemented the code to calculate the intersection of the circle with the line, specifically,
- intersection of Circle2D with Line2D
- intersection of Circle2D with LineSegment2D

It will be nice if we can discuss whether the testcases are sufficient or not.

this task is based on the comment in #186:

> I think it would be useful to have additional methods, like
> * intersection of sphere with line/ray
> * intersection of sphere with plane
> * intersection of sphere with another sphere
> * functions which check if a point lies inside/on/outside of the circle

_Originally posted by @jkalias in https://github.com/mathnet/mathnet-spatial/pull/186#discussion_r1146814522_


src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +134,63 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
             return new Circle2D(center, radius);            
         }
 
+        /// <summary>
+        /// Returns intersection a point2D array if this circle and the given line have the intersections
```
jkalias, line 138: I think this comment could be improved a bit.

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +134,63 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+                .ToArray();
+            return result;
+        }
+
+        private double[] findParameterTs(Line2D line)
+        {
+            // These 2 equations in vector form can be described
+            // (p-cc)^2=r^2 (eq1)
+            // p=s+t*d     (eq2)
+            // , where p is the point on the line and/or circle,
```
jkalias, line 156: _where p is the point on the line, d is the line direction_

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +134,63 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+            // cc is the center of the circle and
+            // r is the radius of the circle.
+            // Substituting (eq2) into (eq1) yields:
+            // ((s+t*d)-c)^2=r^2 (eq3)
```
jkalias, line 160: // ((s+t*d)-cc)^2=r^2 (eq3)

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +134,63 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+            // (eq3) reduces to the following quadratic equation: a*t^2 + b*t + c==0
+
+            var cc = this.Center.ToVector2D(); //center of circle
+            var s = line.StartPoint.ToVector2D();
+            var d = line.Direction;
+            var r = this.Radius;
+
+            var a = d.DotProduct(d);
```
jkalias, line 168: Isn't the line direction `d` a unit vector? Then `a` is always 1.0

src/Spatial/Euclidean/Line2D.cs
```diff
@@ -27,6 +27,21 @@ public struct Line2D : IEquatable<Line2D>, IXmlSerializable
         /// </summary>
         public Point2D EndPoint { get; private set; }
 
+        /// <summary>
```
jkalias, line 30: All these parameters are redundant and are not needed in order to calculate the distance to another 2D point

src/Spatial/Euclidean/Line2D.cs
```diff
@@ -294,5 +328,26 @@ void IXmlSerializable.WriteXml(XmlWriter writer)
             writer.WriteElement("StartPoint", StartPoint);
             writer.WriteElement("EndPoint", EndPoint);
         }
+
+        /// <summary>
+        /// Compute the signed distance from this `line` to the given point `p`. If the sign of the distance is positive, the normal vector from the line is in the same side to the point. 
+        /// </summary>
+        /// <param name="p"></param>
+        /// <returns></returns>
+        public double SignedDistanceTo(Point2D p)
```
jkalias, line 337: Using the notation you employed above for the intersection case, the line is already represented as `q=s+t*d`, where `s` is the start point and `d` is the (unit) direction vector (both of these are cartesian 2D vectors).

Then the distance of any point `p` to this line `q` can be expressed as the norm of this vector `s - (p-s)*d`

src/Spatial/Extensions/DoubleExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Data;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MathNet.Spatial.Extensions
+{
+    internal static class DoubleExtensions
+    {
+        internal static bool IsNearlyEqualTo(this double d, double reference, double tolerance)
```
jkalias, line 12: This is not being used anywhere. If so, it should be removed.

jkalias (changes requested):

src/Spatial/Euclidean/Circle2D.cs
f-frhs: Thanks for the comment.
I'll fix this.

src/Spatial/Euclidean/Circle2D.cs
f-frhs: > Isn't the line direction `d` a unit vector? Then `a` is always 1.0

Yes.
I'll fix this.

src/Spatial/Euclidean/Line2D.cs
f-frhs: Thanks for the comment.
Some might think like you.
I'll remove these parameters.

src/Spatial/Euclidean/Line2D.cs
f-frhs: Thanks for the comment.

If I understand your comment correctly, it seems to me that `s - (p-s)*d` does not mean the vector from the point `p` to the point on the line `q`. 
Because `(p-s)` is the vector.
The operator `*` is for the 2 vectors `(p-s)` and `d`, I cannot understand in this context.

Should I use the normal vector of the line `q`?

src/Spatial/Extensions/DoubleExtensions.cs
f-frhs: Thanks for the comment.
I agree with your opinion, that is, I will remove this.

src/Spatial/Euclidean/Circle2D.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/c2612b61fa2c6f705522c5926f379ec404a77ca4

FYI, there is `c` as it was.
Replacing `c` to `cc` is another problem.

src/Spatial/Euclidean/Circle2D.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/f3e4731f6b5a200841ca36fc8809ced68e479cea

src/Spatial/Euclidean/Line2D.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/847cdaf17e4784a1791d5dd11ddb8f398111cf18

src/Spatial/Extensions/DoubleExtensions.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/2d222ac60492d5ca6a2028cb6a2c7a25193ce640

src/Spatial/Euclidean/Circle2D.cs
jkalias: I was thinking something like "Returns the intersections of this circle with the given line."

src/Spatial/Euclidean/Line2D.cs
jkalias: Actually my original comment war not correct, the correct one should be `s - [(p-s)*d] * d`.
`p - s` is a vector, `(p-s)*d` is a dot product of two vectors (a real number, and then `[(p-s)*d] * d` is again a vector

jkalias:
You mention some commits, but they are in your own branch, not in this PR. So I can't really tell how they are fixed

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +133,66 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+            var cc = this.Center.ToVector2D(); //center of circle
+            var s = line.StartPoint.ToVector2D();
+            var d = line.Direction;
+            var r = this.Radius;
+
+            var a = 1d;
+            var b = 2 * (s.DotProduct(d) - d.DotProduct(cc));
+            var c = (s - cc).DotProduct(s - cc) - r * r;
+
+            var soluions = FindRoots.Polynomial(new[] { c, b, a });
```
jkalias, line 174: `solutions`

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +133,66 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+            var ts = soluions
+                .Where(z => z.IsReal())
+                .Select(z => z.Real)
+                .ToArray();
+            return ts;
+        }
+
+        /// <summary>
+        /// Returns intersection a point2D array if this circle and the given line have the intersections
```
jkalias, line 183: Same comment as above

jkalias (changes requested):

src/Spatial/Euclidean/Circle2D.cs
f-frhs: Much better. Simple, precise and clean.
Thank you.

src/Spatial/Euclidean/Circle2D.cs
f-frhs: Thanks for the comment.
I'm on finxing this typo.

src/Spatial/Euclidean/Circle2D.cs
jkalias: More or less yes, but here you check additionaly that the intersection lies within the line segment. So something like "Returns the intersections of this circle with the given line segment, which lie within the segment."

src/Spatial.Tests/Euclidean/Circle2DTests.cs
```diff
@@ -58,5 +58,34 @@ public void CircleFromThreePointsArgumentException()
 
             Assert.Throws<ArgumentException>(() => { Circle2D.FromPoints(p1, p2, p3); });
         }
+
+        [TestCase("0,0", 1.41421356 /*=sqrt(2)*/, "-1,-1", "+1,+1", "+1,+1", "-1,-1")]
```
jkalias, line 62: I would add some more tests, for which the start/end points of the line do not lie on the circle, just to make sure that we are not running on any edge case.

In addition there are test cases missing for no intersections as well as identical intersections (i.e. line tangent to the circle)

src/Spatial.Tests/Euclidean/Circle2DTests.cs
```diff
@@ -58,5 +58,34 @@ public void CircleFromThreePointsArgumentException()
...
+
+            var expected = new[] { Point2D.Parse(esp0), Point2D.Parse(esp1) };
+            AssertGeometry.AreEqual(actual[0], expected[0]);
+            AssertGeometry.AreEqual(actual[1], expected[1]);
+        }
+
+        [TestCase("0,0", 1.41421356 /*=sqrt(2)*/, "0,0", "+1,+1", "+1,+1")]
+        [TestCase("0,0", 1, "0,0", "+1,0", "+1,0")]
+        [TestCase("0,0", 1, "0,0", "0,+1", "0,+1")]
+        public void CircleIntersectWithLineSegment2D(string sCenter, double radius, string sStart, string sEnd, string sExpected)
```
jkalias, line 80: I would add some more tests, for which the start/end points of the line segment do not lie on the circle, just to make sure that we are not running on any edge case.

In addition there are test cases missing for the following cases:
1. no intersections
2. single intersection with segment whose corresponding line has two intersections
3. no intersection with segment whose corresponding line has two intersections or is tangent on circle
4. identical intersections for line segment on circle (i.e. tangent on circle)

jkalias (changes requested):

src/Spatial/Euclidean/Circle2D.cs
jkalias: But still not changed in the PR 😄

src/Spatial/Euclidean/Line2D.cs
f-frhs: > The more I think about it, the less I think that the notion of "signed" distance of point from a line is something meaningful.

True. You can say that again.

> I think we should provide only one method for the distance, which returns either 0 (point on the line) or a positive value (point not on the line).

From now on, I will call the method `AbsoluteDistance`.
If you don't mind, I will try to implement `AbsoluteDistance` method.
Because, if there are few user stories to use `Line2D.SignedDistance(point)`, 
implementing just `AbsoluteDistance` looks make sense. 
If you mind, please let me know to change my mind.


> I am not sure what you mean exactly by "tangential component", I would call this "the normal component w.r.t. the line"(snip)

Sorry for making you confused by the figure violating the naming convention.
I should have shown you the figure showing the interface between 2 regions like this:
(The figure bellow was taken from https://engineering.purdue.edu/wcchew/ece604f19/EMFTAll20191204.pdf p.31 4.3.1 Faraday's law)

image

For this issue, I arranged this figure imaginally so that the interface is a straight line.

src/Spatial/Euclidean/Circle2D.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/9e33c1b1daff669f22e59ed4051cf230c21f66f6

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: Thanks for the comment about test cases.
This comment looks very similar to the bellow one, can I unify these 2 comments?

src/Spatial/Euclidean/Line2D.cs
jkalias: > From now on, I will call the method `AbsoluteDistance`. If you don't mind, I will try to implement `AbsoluteDistance` method. Because, if there are few user stories to use `Line2D.SignedDistance(point)`, implementing just `AbsoluteDistance` looks make sense. If you mind, please let me know to change my mind.

I think `line.DistanceFrom(p)` of `line.DistanceTo(p)` is straight forward. The `Absolute` does not any value to the function itself, I would argue that it adds confusion.

> Sorry for making you confused by the figure violating the naming convention. I should have shown you the figure showing the interface between 2 regions like this

Stil though, the distance of the line from the point is along the normal direction `n`, and not along the tangential component `t`.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
jkalias: You are right that they are very similar, however the line segment case requires further testing cases, since the intersection points (with the underlying line) might lie outside of the segment itself.

jkalias (commented):
Naming

src/Spatial/Euclidean/Line2D.cs
f-frhs: > I think line.DistanceFrom(p) of line.DistanceTo(p) is straight forward. The Absolute does not add any value to the function itself, I would argue that it adds confusion.

Thank you for arguing this.
Calling `AbsoluteDistance` help us go ahead with this issue.
Your opinion is reasonable.
Is `line.DistanceTo(p)` fine for you?
Either is fine, but this one for now.



> > Sorry for making you confused by the figure violating the naming convention. I should have shown you the figure showing the interface between 2 regions like this
> 
> Stil though, the distance of the line from the point is along the normal direction `n`, and not along the tangential component `t`.

You have a point.
That is what I wanted to say.

Your original comment was:

> Then the distance of any point p to this line q can be expressed as the norm of this vector `s - (p-s)*d`

Your revised comment was:

> Actually my original comment war not correct, the correct one should be `s - [(p-s)*d] * d`.

In my understanding, `s - [(p-s)*d] * d` is along the tangental component `t` (as I quoted the top part of this comment), and I think it is different from the vector which we wanted when we calculate the distance from the line to the given point `p`. We need another expression other than ones you have provided.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: I tried to follow your suggetion with regard to the nubmer of intersections.

Done in the commit https://github.com/f-frhs/mathnet-spatial/commit/b27fdb183d8fe6f3526ebece8c29a90ef3981a09.

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +133,66 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+            // Substituting (eq2) into (eq1) yields:
+            // ((s+t*d)-cc)^2=r^2 (eq3)
+            // (eq3) reduces to the following quadratic equation: a*t^2 + b*t + c==0
+
+            var cc = this.Center.ToVector2D(); //center of circle
+            var s = line.StartPoint.ToVector2D();
+            var d = line.Direction;
+            var r = this.Radius;
+
+            var a = 1d;
```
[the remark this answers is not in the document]
f-frhs: Could you tell me what the difference is between `var a = 1.0` and `var a = 1d`?
If this repo has coding conventions, I failed to read the doc. I will read it and fix this.

src/Spatial/Euclidean/Circle2D.cs
```diff
@@ -131,6 +133,66 @@ public static Circle2D FromPoints(Point2D pointA, Point2D pointB, Point2D pointC
...
+                .ToArray();
+            return ts;
+        }
+
+        /// <summary>
+        /// Returns the intersections of this circle with the given line segment, which lie within the segment.
+        /// </summary>
+        /// <param name="line">the given line-segment</param>
+        /// <returns>intersections as a Point2D Array, depending on the count.</returns>
+        public Point2D[] IntersectWith(LineSegment2D line)
```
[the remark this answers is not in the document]
f-frhs: Thanks for the comment.
Done in the commit https://github.com/f-frhs/mathnet-spatial/commit/5930b12914da065e776a033485fb6d65a05b60ca.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
```diff
@@ -58,5 +59,51 @@ public void CircleFromThreePointsArgumentException()
 
             Assert.Throws<ArgumentException>(() => { Circle2D.FromPoints(p1, p2, p3); });
         }
+
+        [TestCase("0,0", 1, "-10,+10", "+10,+10", 0)]
+        [TestCase("0,0", 1, "-10,+1", "+10,+1", 1)]
+        [TestCase("0,0", 1, "-10,0", "+10,0", 2)]
+        [TestCase("0,0", 1, "-10,-1", "+10,-1", 1 )]
+        [TestCase("0,0", 1, "-10,-10", "+10,-10", 0)]
```
jkalias, line 67: These tests are all dealing with lines which are parallel to the X axis. We are missing other cases (line parallel to Y for example, and the general case)

src/Spatial.Tests/Euclidean/Circle2DTests.cs
```diff
@@ -58,5 +59,51 @@ public void CircleFromThreePointsArgumentException()
...
+        //segments cross the circle's contour just 1 time
+        [TestCase("0,0", 1, "+0,+10", "+10,+10", 0)]
+        [TestCase("0,0", 1, "+0,+1", "+10,+1", 1)]
+        [TestCase("0,0", 1, "+0,0", "+10,0", 1)]
+        [TestCase("0,0", 1, "+0,-1", "+10,-1", 1)]
+        [TestCase("0,0", 1, "+0,-10", "+10,-10", 0)]
+        //segment contains no intersections(px of the startingPoint is too big to intersect with the circle)
+        [TestCase("0,0", 1, "+10,+10", "+100,+10", 0)]
+        [TestCase("0,0", 1, "+10,+01", "+100,+1", 0)]
```
jkalias, line 95: +01?

src/Spatial.Tests/Euclidean/Circle2DTests.cs
```diff
@@ -58,5 +59,51 @@ public void CircleFromThreePointsArgumentException()
...
+            var actual = circle.IntersectWith(line);
+
+            Assert.That(actual.Count(), Is.EqualTo(expectedNumberOfIntersections));
+            //TODO: the intersection should be on the circle
+            Assert.That(actual.All(p => Math.Abs(circle.Center.DistanceTo(p) - circle.Radius) < 1e-6), Is.EqualTo(true), "distance between center and intersection");
+            //TODO: the intersection should be on the line
+        }
+
+        //segment contains the all intersections(same to the cases of circle and line)
+        [TestCase("0,0", 1, "-10,+10", "+10,+10", 0)]
```
jkalias, line 82: Same comment here as in the test above (line always parallel to X)

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: Which part of "+01" di you point out, the center "+10,+01" or the right "+100,+1"?

src/Spatial.Tests/Euclidean/Circle2DTests.cs
```diff
@@ -58,5 +59,51 @@ public void CircleFromThreePointsArgumentException()
...
+        [TestCase("0,0", 1, "+0,+10", "+10,+10", 0)]
+        [TestCase("0,0", 1, "+0,+1", "+10,+1", 1)]
+        [TestCase("0,0", 1, "+0,0", "+10,0", 1)]
+        [TestCase("0,0", 1, "+0,-1", "+10,-1", 1)]
+        [TestCase("0,0", 1, "+0,-10", "+10,-10", 0)]
+        //segment contains no intersections(px of the startingPoint is too big to intersect with the circle)
+        [TestCase("0,0", 1, "+10,+10", "+100,+10", 0)]
+        [TestCase("0,0", 1, "+10,+01", "+100,+1", 0)]
+        [TestCase("0,0", 1, "+10,+00", "+100,0", 0)]
+        [TestCase("0,0", 1, "+10,-01", "+100,-1", 0)]
```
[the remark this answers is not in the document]
f-frhs: ditto

src/Spatial/Euclidean/Circle2D.cs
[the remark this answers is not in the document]
f-frhs: Thanks for the comment.
I respect your feeling, and I will fix this.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: Thanks for the comment.
I will add tests(line parallel to Y for example, and the general case).

src/Spatial.Tests/Euclidean/Circle2DTests.cs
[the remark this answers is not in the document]
jkalias: The center

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/3d3748bca4569c6d5a7aed16c1c3957a1da9e8ff.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/6b8c3db4922851adcba2304d2d81aca9b60a2a63

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/3d3748bca4569c6d5a7aed16c1c3957a1da9e8ff

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/f3815edfaac3955fbd2f08a14fd2280aac9d84bc.

I'm not sure that I understood completely what you mean.
If you find any flaw, please let me know.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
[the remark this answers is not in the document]
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/f3815edfaac3955fbd2f08a14fd2280aac9d84bc.

I'm not sure that I understood completely what you mean.
If you find any flaw, please let me know.

src/Spatial.Tests/Euclidean/Circle2DTests.cs
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/3d3748bca4569c6d5a7aed16c1c3957a1da9e8ff

src/Spatial/Euclidean/Circle2D.cs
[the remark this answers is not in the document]
f-frhs: done in the commit https://github.com/f-frhs/mathnet-spatial/commit/0e7eb693efa625b138e4a86bc2e74674fcc851ad

f-frhs:
Could you tell me which requrest remains?
I lost myself in this PR.

jkalias:
Thank you for your effort and time!

f-frhs:
Thank YOU for your patience.
It was a nice time for me.

PR #238, opened 2023-12-25
Feat: Line2D.DistanceTo(Point2D p)

f-frhs:
I implemented `Line2D.DistanceTo(Point2D p)` which calculates the distance from `this` line to the given point `p`.
This method should always return the value of 0 or positive.

It will be nice if we can discuss whether the testcases is sufficient.

### DIFF
--- a/src/Spatial.Tests/Euclidean/Line2DTests.cs
+++ b/src/Spatial.Tests/Euclidean/Line2DTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using MathNet.Spatial.Euclidean;
 using MathNet.Spatial.Units;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


src/Spatial.Tests/Euclidean/Line2DTests.cs
```diff
@@ -56,6 +56,20 @@ public void LineDirection(string p1s, string p2s, string exs)
             AssertGeometry.AreEqual(ex, line.Direction);
         }
 
+        [TestCase("-1,+1", "+1,+1", "0,0", 1.0)]
+        [TestCase("-1,+1", "+1,+1", "0,3", 2.0)]
+        [TestCase("-1,0", "0,+1", "0,0", 0.70710678)]
+        [TestCase("+1,-1", "+1,+1", "0,0", 1.0)]
+        [TestCase("+1,-1", "+1,+1", "3,0", 2.0)]
+        public void DistanceFromLineToPoint(string p1s, string p2s, string ps, double expectedDistance)
```
jkalias, line 75: I would add other test cases, for which the expected distance is 0, i.e. the line passes through the point.

I would also add some tests for which the order of the start/end line points is reversed, but the expected distance doesn't change.

jkalias (changes requested):

src/Spatial.Tests/Euclidean/Line2DTests.cs
f-frhs: Thanks for your comment.

The test cases you described are meaningful.
I'll do it, thanks.

f-frhs: > I would add other test cases, for which the expected distance is 0, i.e. the line passes through the point.

Done in the commit https://github.com/mathnet/mathnet-spatial/pull/238/commits/ffba33569664f732f81f8c4f1452d42297fd7466.

f-frhs: > I would also add some tests for which the order of the start/end line points is reversed, but the expected distance doesn't change.

Done in the commit https://github.com/mathnet/mathnet-spatial/pull/238/commits/00ce8a6e42ca608cbb126efc31b468d2d3aea84f.

src/Spatial.Tests/Euclidean/Line2DTests.cs
```diff
@@ -56,6 +56,38 @@ public void LineDirection(string p1s, string p2s, string exs)
...
+            new object[] { "-1,+1", "+1,+1", "0,3", 2.0 },
+            new object[] { "-1,+1", "+1,+1", "0,1", 0.0 },
+            new object[] { "-1,0", "0,+1", "0,0", 0.70710678 },
+            new object[] { "-1,0", "0,+1", "-0.5,0.5", 0.0 },
+            new object[] { "+1,-1", "+1,+1", "0,0", 1.0 },
+            new object[] { "+1,-1", "+1,+1", "1,0", 0.0 },
+            new object[] { "+1,-1", "+1,+1", "3,0", 2.0 },
+        };
+
+        [TestCaseSource(nameof(DistanceTestCases))]
```
jkalias: I think the previous structure with `TestCase` was much better. My suggestion with the swapped scenarios was meant to be handled in the `TestCase` definition, not as a separate test function.

jkalias (changes requested):

src/Spatial.Tests/Euclidean/Line2DTests.cs
f-frhs: Thanks for the comment.
Could you tell me why using `TestCode` is better than using `TestCodeSource` for this test?

jkalias: I find the whole `new object[]` a bit overkill, that's it. I guess it's more a matter of personal taste.

f-frhs: Thanks for the comment.

> it's more a matter of personal taste.

Could you show the reason?

jkalias: The reason for what exactly? I prefer compile-time checks and try to avoid `object` as much as possible

f-frhs: Thanks for telling the reasons.

Done in the commit https://github.com/mathnet/mathnet-spatial/pull/238/commits/cd80cf195a07cb3a79f0ccf35dad3f26f14cc89e.
I changed the test-code as you required. I'm not sure that the revised test-code helps the reader understand at a glance that parameters are kept intact except its order.

jkalias: Thank you. This is not what I meant, we can augment the existing tests with a reversed line
```
var line = new Line2D(Point2D.Parse(p1s), Point2D.Parse(p2s));
// existing test

var reversedLine = new Line2D(Point2D.Parse(p2s), Point2D.Parse(p1s));
// distance test with reversed line
```

This way you keep the `TestCase` cases at a minimum

f-frhs: Thanks for showing the example closely.

It is not recommended to test 2 things in a single unit-test.
Or do you want to say that these 2 tests are equivalent because of the symmetry?

jkalias: > It is not recommended to test 2 things in a single unit-test. 

Fully agree

> Or do you want to say that these 2 tests are equivalent because of the symmetry?

Basically yes

src/Spatial/Euclidean/Line2D.cs
```diff
@@ -123,6 +123,19 @@ public static Line2D Parse(string startPointString, string endPointString)
             return new Line2D(Point2D.Parse(startPointString), Point2D.Parse(endPointString));
         }
 
+        /// <summary>
+        /// Returns the straight line Distance to the given point.
```
jkalias, line 127: I would describe this  as 'Returns the length of the perpendicular line to the given point.'

jkalias (approved):

PR #244, opened 2025-05-02
Dotnet tools update

Thorium:
Just a minor update


jkalias:
Have you tried this change for Mac or Linux?

Thorium:
I think the dotnet tool of fake-cli is not even used (yet), the build script still calls the old FAKE.exe and not dotnet tool. Should be updated. 

The paket on the other hand is used and I don't think the old paket even works anymore because there are so many dotnet versions in reference assembly components it doesn't recognise.

Thorium:
Close this for now, as updating to Fake 6 would require a larger refactor for the build.fscx

